PR #25, opened 2019-02-26
Update readme.md

aarmora:
2captcha provides an apiKey instead of a token. Can be confusing to a new user.

Also important to mention that funds will need to be added to 2captcha. 

This should clarify the steps needed to take to make sure this plugin works so users will know this will not work if they haven't added funds and won't think it's the code not working.

### DIFF
--- a/packages/puppeteer-extra-plugin-recaptcha/readme.md
+++ b/packages/puppeteer-extra-plugin-recaptcha/readme.md
@@ -29,13 +29,13 @@ The plugin essentially provides a mighty `page.solveRecaptchas()` method that do
 // it augments the installed puppeteer with plugin functionality
 const puppeteer = require('puppeteer-extra')
 
-// add recaptcha plugin and provide it your 2captcha token
+// add recaptcha plugin and provide it your 2captcha token (= 2captcha apiKey)
 // 2captcha is the builtin solution provider but others work as well.
 // Funds will need to be added to your 2captcha account in order for the 2captcha provider to work
 const RecaptchaPlugin = require('puppeteer-extra-plugin-recaptcha')
 puppeteer.use(
   RecaptchaPlugin({
-    provider: { id: '2captcha', apiKey: 'XXXXXXX' },
+    provider: { id: '2captcha', token: 'XXXXXXX' },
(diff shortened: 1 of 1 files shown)


packages/puppeteer-extra-plugin-recaptcha/readme.md
```diff
@@ -31,10 +31,11 @@ const puppeteer = require('puppeteer-extra')
 
 // add recaptcha plugin and provide it your 2captcha token
```
berstend: Maybe change this comment from 
```js
// add recaptcha plugin and provide it your 2captcha token
```
to
```js
// add recaptcha plugin and provide it your 2captcha token (= 2captcha apiKey)
```

packages/puppeteer-extra-plugin-recaptcha/readme.md
```diff
@@ -31,10 +31,11 @@ const puppeteer = require('puppeteer-extra')
...
 // 2captcha is the builtin solution provider but others work as well.
+// Funds will need to be added to your 2captcha account in order for the 2captcha provider to work
 const RecaptchaPlugin = require('puppeteer-extra-plugin-recaptcha')
 puppeteer.use(
   RecaptchaPlugin({
-    provider: { id: '2captcha', token: 'XXXXXXX' },
+    provider: { id: '2captcha', apiKey: 'XXXXXXX' },
```
berstend: We need to change this back to `token` as we'd otherwise need to update the code as well.

aarmora: Good call! My bad for not thinking of that.

packages/puppeteer-extra-plugin-recaptcha/readme.md
aarmora: Good thinking.

aarmora:
@berstend I've not used github for PRs for a while. I think it's all corrected now. Let me know if there is more I need to do.

berstend:
@aarmora apologies for the delay on this - I did a larger refactor and added your changes as well (#81). :) 

I hope you don't mind if I don't merge this PR - it was easier for me to do it while refactoring. Thanks for contributing! 👍

aarmora:
@berstend No problem at all, of course!

PR #37, opened 2019-05-03
Fix getParameter fallback usage

Palujust:
I don't think the reference to the original getParameter function works correctly. I've found a specific case where the website I'm trying to hit uses WebGL as part of its browser fingerprinting. The JavaScript tries to get several parameters (that do not have hardcoded values here) and Chrome spits out `TypeError: getParameter is not a function`. Adding the old version of the function to the prototype and calling it that way seems to work.

You can replicate the existing bug by loading this plugin, navigating to any web page and running the following in the console:

```
const canvas = document.createElement('canvas');
const context = canvas.getContext('webgl');
context.getParameter(123);
```

My change allows this code to run correctly and return correct values for any Parameter that isn't over-written by the plugin.

### DIFF
--- a/packages/puppeteer-extra-plugin-stealth/evasions/webgl.vendor/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/webgl.vendor/index.js
@@ -19,7 +19,7 @@ class Plugin extends PuppeteerExtraPlugin {
     await page.evaluateOnNewDocument(() => {
       try {
         /* global WebGLRenderingContext */
-        const getParameter = WebGLRenderingContext.getParameter
+        WebGLRenderingContext.prototype.oldGetParameter = WebGLRenderingContext.prototype.getParameter;
         WebGLRenderingContext.prototype.getParameter = function (parameter) {
           // UNMASKED_VENDOR_WEBGL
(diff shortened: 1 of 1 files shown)


dcramer:
Can confirm. kmart.com exhibits this behavior.

berstend:
Thanks for the PR @Palujust 👍 I'm currently a bit busy refactoring the core, will look into this as soon as I find the time :)

berstend:
@Palujust I made a massive update to the stealth plugin, which fixes all current issues:
https://github.com/berstend/puppeteer-extra/tree/master/packages/puppeteer-extra-plugin-stealth#v240

I've also added before/after tests to make this whole thing less voodoo - would you mind rebasing this PR and adding a before/after test for the canvas behaviour? Without those tests it's tough for me to merge in new fixes. :)

Thanks a lot for your contribution!

berstend:
Unfortunately there's more broken with this, e.g. `context.getParameter.toString()` will not be `function getParameter() { [native code] }`.

Looking into a better way now.

berstend:
Fixed in `puppeteer-extra-plugin-stealth: 2.4.3`

Issue #27, opened 2019-03-14
How to use with puppeteer-core?


mrgodhani:
HI,

I don't want to install full puppeteer package as it also installs a browser with it. If I want to use puppeteer-core and use installed browser to connect using chrome-launcher how would I use it with puppeteer-extra?

Jumblemuddle:
This would be great; at the moment I have to manually symlink my chromium installation into my node_modules in order to use it, because I can't use puppeteer-core along with puppeteer-extra.

haggholm:
I’ve opened https://github.com/berstend/puppeteer-extra/pull/43, so we’ll see what happens; in a lot of solutions, you can probably use the new npm alias functionality (`npm install puppeteer@npm:puppeteer-core`) (https://github.com/npm/rfcs/blob/latest/implemented/0001-package-aliases.md)

berstend:
> according to ["puppeteer vs puppeteer-core"](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteer-vs-puppeteer-core) [](/GoogleChrome/puppeteer/blob/HEAD@{2019-09-29T16:37:13Z}/docs/api.md#puppeteer-vs-puppeteer-core) they have two main differences...
> 
> > To sum up, the only differences between puppeteer-core and puppeteer are:
> > 
> > * puppeteer-core doesn't automatically download Chromium when installed.
> > * puppeteer-core ignores all PUPPETEER_* env variables.
> 
> You can avoid downloading chrome using the [`PUPPETEER_SKIP_CHROMIUM_DOWNLOAD`](https://github.com/GoogleChrome/puppeteer/blob/v0.10.2/docs/api.md#environment-variables) [](/GoogleChrome/puppeteer/blob/HEAD@{2019-09-29T16:37:13Z}/docs/api.md#environment-variables) environment variable during installation. This makes `puppeteer` almost identical to `puppeteer-core` with the added benefit of convenient env vars.
> 
> e.g. `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1 npm install puppeteer`

Closing this for now, please re-open if the above solution doesn't work in all scenarios. :)

PR #43, opened 2019-06-05
Try to fallback to `puppeteer-core` if `puppeteer` is not available

haggholm:
What it says on the label. Maybe not the flexible as it doesn’t make it easier to dropin forks &c., but it’s a quick solution that probably does what *most* people are looking for. Cf. https://github.com/berstend/puppeteer-extra/issues/27

### DIFF
--- a/packages/puppeteer-extra/index.js
+++ b/packages/puppeteer-extra/index.js
@@ -3,7 +3,11 @@
 let Puppeteer
 try {
   // https://github.com/GoogleChrome/puppeteer/pull/3208
-  Puppeteer = require('puppeteer')
+  try {
+    Puppeteer = require('puppeteer')
+  } catch (err) {
+    Puppeteer = require('puppeteer-core')
+  }
 } catch (err) {
   console.warn(`
     Puppeteer is missing. :-)


shirshak55:
Try catch is really not the good way to handle such flow?

Kikobeats:
You in case you are interested, I created a module for that purpose https://github.com/Kikobeats/require-one-of 😂

berstend:
@haggholm apologies for the delay and thanks for contributing :) I'm currently refactoring the core to TS and will make sure to use your fix if it still seems like the best way in the new setup.

berstend:
Fix for the time being:

> according to ["puppeteer vs puppeteer-core"](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteer-vs-puppeteer-core) [](/GoogleChrome/puppeteer/blob/HEAD@{2019-09-29T16:37:13Z}/docs/api.md#puppeteer-vs-puppeteer-core) they have two main differences...
> 
> > To sum up, the only differences between puppeteer-core and puppeteer are:
> > 
> > * puppeteer-core doesn't automatically download Chromium when installed.
> > * puppeteer-core ignores all PUPPETEER_* env variables.
> 
> You can avoid downloading chrome using the [`PUPPETEER_SKIP_CHROMIUM_DOWNLOAD`](https://github.com/GoogleChrome/puppeteer/blob/v0.10.2/docs/api.md#environment-variables) [](/GoogleChrome/puppeteer/blob/HEAD@{2019-09-29T16:37:13Z}/docs/api.md#environment-variables) environment variable during installation. This makes `puppeteer` almost identical to `puppeteer-core` with the added benefit of convenient env vars.
> 
> e.g. `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1 npm install puppeteer`

I'm closing this PR for now but keep this in mind when I refactor the core to TypeScript and will also update the docs. :)

berstend:
`v3.1.3` will now attempt to require `puppeteer-core` as well :)

elyas-bhy:
@berstend is this documented anywhere?

PR #63, opened 2019-09-22
Add puppeteer-extra-plugin-adblocker

remusao:
Hi,

I am the maintainer of [`@cliqz/adblocker-puppeteer`](https://www.npmjs.com/package/@cliqz/adblocker-puppeteer) which allows to block ads and trackers in any puppeteer project. I thought this would be a nice addition to your collection of plugins!

I created a small POC by copying the recaptcha plugin and adapting it for the adblocker plugin. It allows to either block only ads (default) or trackers as well. There is an option which allows to cache the adblocker on disk to avoid having to perform network requests every time it is initialized.

Best,
Rémi

### DIFF
--- a/packages/puppeteer-extra-plugin-adblocker/package.json
+++ b/packages/puppeteer-extra-plugin-adblocker/package.json
@@ -1,0 +1,78 @@
+{
+  "name": "puppeteer-extra-plugin-adblocker",
+  "version": "2.11.0",
+  "description": "A puppeteer-extra plugin to block ads and trackers.",
+  "main": "dist/index.cjs.js",
+  "module": "dist/index.esm.js",
+  "typings": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "repository": "berstend/puppeteer-extra",
+  "homepage": "https://github.com/berstend/puppeteer-extra/tree/master/packages/puppeteer-extra-plugin-adblocker",
+  "author": "remusao",
+  "license": "MIT",
+  "scripts": {
+    "clean": "rm -rf dist/*",
+    "tscheck": "tsc --pretty --noEmit",
+    "build": "yarn clean && tsc --module commonjs && rollup -c rollup.config.ts",
+    "docs": "exit 0",
+    "docs2": "rm -rf docs/* && typedoc --module commonjs --readme none --target ES6 --theme markdown --excludeNotExported --excludeExternals --excludePrivate --out docs --mode file src/index.ts",
+    "test": "TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' ava -v src/**.test.ts",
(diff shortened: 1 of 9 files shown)


shirshak55:
honestly what i do is just load ublock origin and it works perfectly fine. But this is something nice to have.

berstend:
@remusao oh wow, that looks super interesting! I'll need a moment to get things back in order (was swamped with other things and neglected this project a bit) but will make sure to look at this once the dust has settled. Thanks for contributing!

berstend:
@remusao oh wow, you made my day! thanks a lot for contributing this high quality PR and plugin. 👍 

I made some small adjustments before merging & publishing it (hope you don't mind):
- moved `@types/chrome` from dev to regular dependencies (better out of the box experience as they would be otherwise missing)
- updated all deps, including `"@cliqz/adblocker-puppeteer": "^1.4.0"`
- refactored `cacheDir`: cache is now enabled by default and uses a default temp dir, with the option to disable caching or to provide a specific cache dir
- removed the explicit ready() so the wait happens behind the scenes automatically (beforeLaunch, beforeConnect are blocking hooks) :)

I really like this plugin, I made it the default one in main readme (which people tend to just copy&paste) 😄 

Thanks again, nice work!

remusao:
@berstend My pleasure! I'm happy if it can be useful for you and the community! :D Thanks for getting the PR up to speed with latest `master` branch + improvements!

entrptaher:
The puppeteer version on this plugin is hardcoded, which makes yarn install double installing puppeteer even though one is used.

remusao:
> The puppeteer version on this plugin is hardcoded, which makes yarn install double installing puppeteer even though one is used.

Hi @entrptaher,

I am the maintainer of the `@cliqz/adblocker` library, after double-checking it seems that the `electron` dependency is only a [peer dependency](https://github.com/cliqz-oss/adblocker/blob/master/packages/adblocker-electron/package.json#L37) and it only says that the version should be higher than `8`. Which version are you using in your project? Also, how would you suggest this be changed to be more flexible for people using the plugin?

Thanks for your feedback,

entrptaher:
This gets installed when I install the adblocker plugin on this repo. You can see it's hardcoded to version 2.0.0

https://github.com/berstend/puppeteer-extra/blob/master/packages/puppeteer-extra-plugin-adblocker/package.json#L53

remusao:
> This gets installed when I install the adblocker plugin on this repo. You can see it's hardcoded to version 2.0.0.
> 
> https://github.com/berstend/puppeteer-extra/blob/master/packages/puppeteer-extra-plugin-adblocker/package.json#L53
> 
> PS: I frequently install things for development, specially on a monorepo and this is devDependency. Think how waiting for 2 minutes for two puppeteer chrome download feels :D

I see, this seems to be an issue with puppeteer-extra here, I initially thought this might be caused by `@cliqz/adblocker-puppeteer`. I guess @berstend might want to have a look instead then.

Issue #33, opened 2019-04-08
Being detected by Distil Networks


mjfeintuch:
I am trying to automate logon to site that is using Distil. When I attempt to logon, I get a captcha.
![Capture](https://user-images.githubusercontent.com/38329675/55742480-c8161880-59fd-11e9-8f9a-0042b3310b0f.PNG)

petrpatek:
I think it might be related to the typing speed. Have you tried to lower the speed of typing? I would also try to implement random delays between actions like clicking the submit button after filling out the form.

bjesus:
I'm being detected by Distil without even typing anything, just trying to see a website.

shirshak55:
@bjesus did u use puppetter stealth plugin or not. And by the way distill network can use many heuristics to detect bot so I don't think it is easy to find that heuristic.

bjesus:
Hi @shirshak55, yes, I did use the stealth plugin of course, otherwise I wouldn't report it here :)

BorysTyminski:
I have same issue.

shirshak55:
check if real browser has also same issue or not. They may be using ip to check bot activity guys. And did u installed any other extensions, additional fonts etc?

BorysTyminski:
In normal browser it just display their site.

I'm not using any proxy. I didn't install any other packages. I even didn't go to https://www.distilnetworks.com/ programically but manually typing url in puppeteer chromium window. They instantly detect it somehow.

shirshak55:
@BorysTyminski  using chromium?

shirshak55:
And do they detect one time only or each time. Because sometime due to new brand new fresh browser they may be suspicious . And save user data folder so they people u are same user.

BorysTyminski:
Doesn't puppeteer use chromium? I set `headless: false,` and I just paste url and it instantly detected me. I didn't open this site never with puppeteer without stealth plugin so I doubt they saved me as a suspicious user.

shirshak55:
@BorysTyminski we can use chrome. 

And if u are using chromium be sure to change user agent.  which url give me i would like to test if it detects me or not.

BorysTyminski:
https://www.distilnetworks.com/

BorysTyminski:
@shirshak55 did you check it? Do they detect you as well?

shirshak55:
they deected on that website only other not detected. it has to do with
canvas probably finger printing issue

BorysTyminski:
What do you mean with:

[...]

PR #77, opened 2019-11-27
Fix getParameter spoofing for WebGLRenderingContext

petr-vysotskiy:
Hey,

We've faced the issue mentioned here:
https://github.com/berstend/puppeteer-extra/issues/33#issuecomment-497033554

So, here is the fix.

### DIFF
--- a/packages/puppeteer-extra-plugin-stealth/evasions/webgl.vendor/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/webgl.vendor/index.js
@@ -19,7 +19,7 @@ class Plugin extends PuppeteerExtraPlugin {
     await page.evaluateOnNewDocument(() => {
       try {
         /* global WebGLRenderingContext */
-        const getParameter = WebGLRenderingContext.getParameter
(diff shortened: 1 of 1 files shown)


berstend:
@petr-vysotskiy awesome, thanks for the PR! 👍 

Do you know if this breaks any existing behavior or pages or is it safe to merge?
I'm currently in the middle of rewriting the core to TS and can't test this in detail. :)

Thanks a lot!

petr-vysotskiy:
You are welcome. Thanks @berstend for the work you've done, it helped us. We've tested the fix and it works fine with Distil.

petr-vysotskiy:
> But, Distil still detects me as a bot.
Can you see any other errors? Have you used a proxy to make sure your IP isn't blocked by Distil? Which site are you trying to visit using puppeter?

kfunk1:
Unfortunately, I don't see any other error.
I am not using a proxy.
I am always able through a real chrome, so it's not an ip problem.
https://www.axs.com/browse/music

I've tried in headless mode on, it seems to be working.

Issue #91, opened 2019-12-05
Request interception with multiple plugins enabled


ItzRabbs:
I think the underlying issue is similar to https://github.com/berstend/puppeteer-extra/issues/90 but there is another error that comes up and I think applies for more cases.

```
const puppeteer = require('puppeteer-extra');

puppeteer.use(require('puppeteer-extra-plugin-anonymize-ua')());
puppeteer.use(require('puppeteer-extra-plugin-stealth')());
const blockResourcesPlugin = require('puppeteer-extra-plugin-block-resources')({
  blockedTypes: new Set(['stylesheet', 'image', 'media', 'font'])
});
puppeteer.use(blockResourcesPlugin);

(async function () {
  const browser = await puppeteer.launch({ headless: false });
  const page = await browser.newPage();

  await page.goto('https://www.google.com');

  await browser.close();
})();
```

Running the code above produces two errors (but multiple times, presumably for every request).

```
(node:1300) UnhandledPromiseRejectionWarning: Error: You set up a request listener but no interception. If you intend to modify requests you need to add: `await page.setRequestInterception(true)`.
    at Request.continue (I:\CAG Portal\NodeJS\node_modules\puppeteer-extra-plugin-stealth\evasions\accept-language\index.js:120:19)
    at Plugin.onRequest (I:\CAG Portal\NodeJS\node_modules\puppeteer-extra-plugin-block-resources\index.js:101:60)
    at Page.emit (events.js:215:7)
    at NetworkManager.<anonymous> (C:\Users\user\AppData\Roaming\nvm\v9.7.1\node_modules\puppeteer\lib\Page.js:114:68)
    at NetworkManager.emit (events.js:210:5)
    at NetworkManager._onRequest (C:\Users\user\AppData\Roaming\nvm\v9.7.1\node_modules\puppeteer\lib\NetworkManager.js:240:10)
    at NetworkManager._onRequestWillBeSent (C:\Users\user\AppData\Roaming\nvm\v9.7.1\node_modules\puppeteer\lib\NetworkManager.js:173:14)
    at CDPSession.emit (events.js:210:5)
    at CDPSession._onMessage (C:\Users\user\AppData\Roaming\nvm\v9.7.1\node_modules\puppeteer\lib\Connection.js:200:12)
    at Connection._onMessage (C:\Users\user\AppData\Roaming\nvm\v9.7.1\node_modules\puppeteer\lib\Connection.js:112:17)

[...]

PR #104, opened 2019-12-12
feat(plugin-stealth): Add new user-agent-language-platform plugin

Niek:
This plugin basically replaces the following evasions:
https://github.com/berstend/puppeteer-extra/tree/master/packages/puppeteer-extra-plugin-stealth/evasions/accept-language
https://github.com/berstend/puppeteer-extra/tree/master/packages/puppeteer-extra-plugin-stealth/evasions/navigator.languages
At the same time, it fixes https://github.com/berstend/puppeteer-extra/issues/91

Instead of changing request headers and the ``navigator.languages`` (``navigator.language`` was missing by the way) properties as in the evasions above, we use call ``setUserAgentOverride`` in the CDP Network domain directly which is a better way of achieving this.

In addition, this evasion allows setting the ``navigator.platform`` and ``navigator.vendor`` properties, which makes it a lot less easy to detect puppeteer.

### DIFF
--- a/packages/puppeteer-extra-plugin-stealth/evasions/user-agent-language-platform/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/user-agent-language-platform/index.js
@@ -1,0 +1,68 @@
+'use strict'
+
+const { PuppeteerExtraPlugin } = require('puppeteer-extra-plugin')
+
+const userInfo = msg => {
+  console.log(`INFO: puppeteer-extra-plugin-stealth(user-agent-language-platform): ${msg}`)
+}
+
+/**
+ * By default puppeteer will have a fixed locale setting, represented in the `Accept-Language` header and in `navigator.languages`.
+ * In addition, the `navigator.platform` and `navigator.vendor` properties are always set to the host values, so for example platform `Linux` which makes detection very easy.
+ * 
+ * This plugin fixes these issues. Please know that you cannot use the regular ``page.setUserAgent()`` puppeteer call in your code,
+ * as it will reset the language and platform values you set with this plugin.
+ *
+ * @example
+ * const puppeteer = require("puppeteer-extra")
+ *
+ * const StealthPlugin = require("puppeteer-extra-plugin-stealth")
+ * const stealth = StealthPlugin()
+ * // Remove this specific stealth plugin from the default set
+ * stealth.enabledEvasions.delete("user-agent-language-platform")
+ * puppeteer.use(stealth)
+ *
+ * // Stealth plugins are just regular `puppeteer-extra` plugins and can be added as such
+ * const UserAgentLanguagePlatformPlugin = require("puppeteer-extra-plugin-stealth/evasions/user-agent-language-platform")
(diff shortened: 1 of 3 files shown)


berstend:
@Niek Interesting, thanks a lot for this PR!

I wasn't aware that you could modify the `Accept-Language` request header using `Network.setUserAgentOverride`.

Would you mind adding a test file that does before/after tests with this plugin?
You could use e.g. this file as the basis: https://github.com/berstend/puppeteer-extra/blob/master/packages/puppeteer-extra-plugin-stealth/evasions/accept-language/index.test.js

I would do it myself but there's no good way in Github (that I'm aware of) to push commits to this PR (without you using the "Contributor" status). :)

Is there a specific reason to add `navigator.vendor` here as well, as opposed to making a specific plugin for it? I personally lean towards making the evasions as modular as possible and splitting them up if it makes sense.

The rest (userAgent, acceptLanguage, platform) is set by a single `Network.setUserAgentOverride` call, so it make sense to group it into a single plugin.

Thanks again!

berstend:
#91 is still something I need to look into (so end users can use the `adblock` plugin while still being able to do their own request interception), but if this PR fixes the `accept-language` evasion than this would make fixing #91 less critical in my opinion.

berstend:
Once you add a test file you can specifically run it like so (from the root of the `puppeteer-extra-plugin-stealth` folder):

```bash
yarn ava -v ./evasions/webgl.vendor/index.test.js
```

packages/puppeteer-extra-plugin-stealth/evasions/user-agent-language-platform/index.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+  }
+
+  async onPageCreated(page) {
+    this.debug('onPageCreated - Will set these user agent options', {
+      opts: this.opts
+    })
+
+    page._client.send('Network.setUserAgentOverride', {
+      userAgent: this.opts.userAgent || (await page.browser().userAgent()),
```
berstend: Could it make sense to add `.replace('HeadlessChrome/', 'Chrome/')` here as well?

Niek: Yes, good idea! Added now

Niek:
@berstend Thank *you* for the amazing repo! I will add the tests ASAP to the PR.

I have removed the ``navigator.vendor`` part, I agree that would probably be better in a separate plugin.

Niek:
I added the tests, let me know if you need other changes @berstend

berstend:
@Niek awesome, thanks a lot! Will look at the PRs asap. :)

berstend:
@Niek thanks again and apologies for the delay (I'd like to blame the holidays) 😄 

I've removed the `user-agent`, `accept-language` evasions which are made obsolete by this one and renamed it `user-agent-override` (to keep it more concise).

Will merge in some other fixes & PRs and do a release afterwards. :)

Thanks! 👍

berstend:
Released in `puppeteer-extra-plugin-stealth@2.4.7`. :) Thanks again!

Niek:
Awesome, thanks for the great work!

pastuh:
What if i want to randomize User Agent?
Can i somehow turn off user-agent-override but still use StealthPlugin ?

PR #121, opened 2020-01-02
Keep user options when defaults are not defined

jozsi:
Hello,

If the defaults are not defined (or defined as an empty object) in a plugin, user options are ignored.

I've fixed the issue (+ added test to confirm).

Thanks!


jozsi:
CI is now passing after #128

berstend:
@jozsi Nice addition, thanks for contributing! 👍 Also thanks for adding the test, this makes merging this in very easy :)

PR #128, opened 2020-01-05
Fix CI issues

berstend:
WIP


berstend:
This issues is very likely the culprit:
https://github.com/sindresorhus/type-fest/issues/19

PR #130, opened 2020-01-07
Fix for calling functions on proxy (navigator.webdriver evasion)

Niek:
This fixes function calls on the navigator object, for example ``navigator.javaEnabled()`` which is called by Google Analytics.

Related Chromium bug: https://bugs.chromium.org/p/v8/issues/detail?id=5773

### DIFF
--- a/packages/puppeteer-extra-plugin-stealth/evasions/navigator.webdriver/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/navigator.webdriver/index.js
@@ -20,8 +20,12 @@ class Plugin extends PuppeteerExtraPlugin {
       Object.defineProperty(window, 'navigator', {
         value: new Proxy(navigator, {
           has: (target, key) => (key === 'webdriver' ? false : key in target),
-          get: (target, key, receiver) =>
-            key === 'webdriver' ? undefined : target[key]
+          get: (target, key) =>
+            key === 'webdriver'
+              ? undefined
+              : typeof target[key] === 'function'
+              ? target[key].bind(target)
+              : target[key]
         })
       })
     })


berstend:
Interesting. I wonder why that's necessary though - function calls are covered by the `apply` proxy trap? It shouldn't be necessary to bind the target context during `Object.get()` (which the `get` trap is taking care of).

I did some quick search if `Reflect.get` would take care of setting the this context to the target but it doesn't seem to be doing that. I'm a bit confused as of why this is necessary but if it fixes the issues you've mentioned we shall merge it in. :)

berstend:
Ah, well - it's a bug in Chromium so it doesn't need to make sense. 😄

berstend:
I've added a regression test so this won't happen again in the future:

```js
test('stealth: regression: wont kill other navigator methods', async t => {
  const puppeteer = addExtra(vanillaPuppeteer).use(Plugin())
  const browser = await puppeteer.launch({ headless: true })

  try {
    const data = await page.evaluate(() => navigator.javaEnabled())
    t.is(data, undefined)
  } catch (err) {
    t.is(err, undefined)
  }
})
```

https://github.com/berstend/puppeteer-extra/commit/d495042787a7c3443f0431df20324182fbb0b882

Niek:
I agree it's weird, but apparently the apply trap is not called in case of native functions. Maybe it's fixed in a future Chromium release, so the regression test is a good idea.

berstend:
Good find regardless. :) I skimmed the rest of the evasions but this seems to be the only occurrence where this fix is required.

berstend:
Actually that's not true, found some other spots that had this issues and added the fix. 😅

Niek:
Oops... Didn't even think of checking other evasions, good catch!

berstend:
image

Released as `puppeteer-extra-plugin-stealth@2.4.9` - thanks again @Niek! 💪

PR #133, opened 2020-01-08
Fix Flash plugin on Chrome 76+, fix #131

### LINKED ISSUE #131: Flash plugin no longer working
Moved [this comment](https://github.com/berstend/puppeteer-extra/issues/75#issuecomment-566066359) to a new issue for visibility:

Flash support is (no longer) working in at least Chrome 80, sample code:
```javascript
'use strict'

const puppeteer = require('puppeteer-extra')

const userName = require('os').userInfo().username
const pluginVersion = '32.0.0.303'
const pluginPath = `
  /Users/${userName}/Library/Application Support/Google/Chrome/PepperFlash/${pluginVersion}/PepperFlashPlayer.plugin
`.trim()

puppeteer.use(
    require('puppeteer-extra-plugin-flash')({
        pluginPath, // this should not be necessary, but setting just in case
        pluginVersion // the result is the same with or without
    })
);

(async () => {
    const browser = await puppeteer.launch({ headless: false, executablePath: '/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome', })
    await page.goto('http://ultrasounds.com', { waitUntil: 'domcontentloaded' })
})()
```

Result: 
Screen Shot 2019-12-16 at 14 37 13

chrome://version shows the correct Flash plugin/version, so I think it's more of a preference issue - this code no longer seems to work: https://github.com/berstend/puppeteer-extra/blob/491c82d042a8607e41d422b3e6bd8806292a151c/packages/puppeteer-extra-plugin-flash/index.js#L73-L110

Niek:
This fixes issue https://github.com/berstend/puppeteer-extra/issues/131 and at the same time cleans up the preference code a bit. Please test carefully, but works fine on my tests.

Credit to this post by @cztomczak which put me in the right direction: https://magpcss.org/ceforum/viewtopic.php?p=43396&sid=fb0574fa7afface5976d913e330dba0b#p43396

### DIFF
--- a/packages/puppeteer-extra-plugin-flash/index.js
+++ b/packages/puppeteer-extra-plugin-flash/index.js
@@ -72,34 +72,10 @@ class Plugin extends PuppeteerExtraPlugin {
     }
     const allowFlashPreferences = {
       profile: {
-        content_settings: {
-          exceptions: {
-            flash_data: {
-              '*,*': {
-                setting: {
-                  flashPreviouslyChanged: true
-                }
-              }
-            },
-            permission_autoblocking_data: {
-              '*,*': {
-                setting: {
-                  Flash: {
-                    dismiss_count: 1
-                  }
-                }
-              }
-            },
-            plugins: {
-              '*,*': {
-                per_resource: {
-                  'adobe-flash-player': 1
-                }
-              }
-            }
-          },
-          pref_version: 1
-        }
+        managed_default_content_settings: {
+          plugins: 1
+        },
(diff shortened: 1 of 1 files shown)


berstend:
Awesome find @Niek - I did not expect another workaround to be possible, given the recent crackdown on Flash content in Chromium. Thanks for investigating this. 👍

berstend:
Published as `puppeteer-extra-plugin-flash@2.2.5`, thanks!

Issue #156, opened 2020-02-19
[puppeteer-extra-plugin-block-resources] Error: Request is already handled!


ItzRabbs:
This is similar to issue https://github.com/berstend/puppeteer-extra/issues/90 and https://github.com/berstend/puppeteer-extra/issues/91 but is occured sporadically with this being the only plugin enabled and no other request interception exists in the code.

Unfortunately, I'm not allowed to post the code that's running but it's scraping a terribly designed web page with a few dozen iframes that continuously change as the code loops through them. The iframes also use the same resources.

I could be wrong, but I believe this error is occuring because the onRequest function is called multiple times for such "duplicate" resources even though they were handled once before. Regardless, I'm thinking that a solution would be to check if the request has been handled before request.continue() or request.abort() is called, as puppeteer does in the code below.

https://github.com/puppeteer/puppeteer/blob/c2651c2b5cc888ebd0ca6be87063d9f98b9eb59c/lib/NetworkManager.js#L456

```
assert(!this._interceptionHandled, 'Request is already handled!');
```

ItzRabbs:
Pull request: https://github.com/berstend/puppeteer-extra/pull/157

liamcharmer:
Is this really solved?

ItzRabbs:
I did this manually and it solved it for me. Doesn't seem to be merged in with the master.

liamcharmer:
What can i do then to avoid this?? 🌊

benallfree:
FYI I have a Puppeteer PR in to solve this https://github.com/puppeteer/puppeteer/pull/6733

seth100:
> FYI I have a Puppeteer PR in to solve this [puppeteer/puppeteer#6733](https://github.com/puppeteer/puppeteer/pull/6733)

I hope they will accept it as soon as possible!

benallfree:
@seth100 Try the package I just published https://www.npmjs.com/package/enchant-puppeteer

seth100:
> @seth100 Try the package I just published https://www.npmjs.com/package/enchant-puppeteer

thanks @benallfree, but trying to install that I get 404

benallfree:
@seth100 My fault, fixed in 0.0.2. Please install again: `npm i enchant-puppeteer@latest`.

seth100:
@benallfree it works great! No error no more! Thank you

benallfree:
[...]

PR #157, opened 2020-02-19
Check if request is handled before continue() or abort() is called

ItzRabbs:
In reference to issue https://github.com/berstend/puppeteer-extra/issues/156

### DIFF
--- a/packages/puppeteer-extra-plugin-block-resources/index.js
+++ b/packages/puppeteer-extra-plugin-block-resources/index.js
@@ -95,6 +95,7 @@ class Plugin extends PuppeteerExtraPlugin {
    * @private
    */
   onRequest(request) {
+    if (request._interceptionHandled) return
     const type = request.resourceType()
     const shouldBlock = this.blockedTypes.has(type)
     this.debug('onRequest', { type, shouldBlock })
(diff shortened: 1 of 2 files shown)


packages/puppeteer-extra-plugin-block-resources/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "puppeteer-extra-plugin-block-resources",
-  "version": "2.2.4",
+  "version": "2.2.5",
```
brunogaspar, line 3: Can you revert this change please?

packages/puppeteer-extra-plugin-block-resources/index.js
```diff
@@ -111,6 +112,6 @@ class Plugin extends PuppeteerExtraPlugin {
   }
 }
 
-module.exports = function(pluginConfig) {
+module.exports = function (pluginConfig) {
```
brunogaspar, line 115: Can you revert this change please?

There's no strict guidelines on how the code should be linted, therefore it's best to keep it as similar as possible for now :)

berstend:
This PR is one year old, closing in favor of the discussions happening at #156 and #364

PR #164, opened 2020-03-09
fix: Adding explicit declaration of fsevents as an optional dependency in package.json

starrify:
Therefore it may be processed properly when there's `--ignore-optional` (thanks to yarnpkg/yarn#7273).
Saves GNU/Linux users from the incompatible platform issue during installation.
Also regenerated `yarn.lock` after the change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,9 @@
     "prettier": "^1.19.1",
     "prettier-config-standard": "^1.0.1"
   },
+  "optionalDependencies": {
+    "fsevents": "^2.1.2"
+  },
   "resolutions": {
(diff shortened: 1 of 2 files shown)


starrify:
:point_up: I've got a quick look at the failed build jobs and they don't look really like errors that are caused by my change.

All failures happen in the `iframe.contentWindow` evasion of `puppeteer-extra-plugin-stealth` are observed to be related to `hasRecaptchaPopup`.
Hypothesis: The page may somehow fail to load properly in one second (the test case has `page.waitFor(1000)`), or there may be no recaptcha challenge at all (the IP address was good enough, etc.).

PR #178, opened 2020-04-02
Bypass optional parameters to all plugins and deps

avchu:


avchu:
I found more beautiful solution )

PR #183, opened 2020-04-13
feat(plugin-repl): add definition file

peterblazejewicz:
This adds definition file to be shipped with the package itself.
Useful for TS usage, when JS code is checked for errors with TSC.

Thanks!


berstend:
Published as `puppeteer-extra-plugin-repl@2.2.7`

PR #197, opened 2020-05-12
Enhanced testing for the user-agent evasion

starrify:


berstend:
Published as `puppeteer-extra-plugin-stealth@2.4.12`, thanks @starrify 👍

PR #198, opened 2020-05-12
plugin-stealth: Do not apply the default values twice for the user-agent evasion, making it possible for users not to introduce specific overrides

starrify:


jozsi:
Can we detail this a bit? I remember removing the `user-agent-override` evasion from the stealth plugin for some reason, can't remember why - wonder if this would fix it.

starrify:
Hi @jozsi , thanks for asking! The motivation is to allow users to override only specific values (e.g. only `acceptLanguage`) while keep the rest unmodified.

berstend:
The `user-agent-override` plugin needs some cleanup, e.g. we're currently setting the platform to `Win32` by default but not modifying the User-Agent in case it contains Linux 😄 

I think we should still mask Linux usage by default, while allowing disabling any modifications here? We could e.g. support passing `null` values and in that case not modify this value at all.

Issue #228, opened 2020-06-18
Support puppeteer 5


Niek:
Puppeteer 4  (*edit*: 5 now) has been released. The changes are quite trivial for puppeteer-extra, I think the most important is to remove this:
https://github.com/berstend/puppeteer-extra/blob/a2a06f971973067a631b1156cf826a034b1323ee/packages/puppeteer-extra-plugin-stealth/index.js#L155

I can do a PR but I'm not sure if this repo is maintained anymore @berstend?

brunogaspar:
Author is MIA since end of February, so i guess sending a pull request is just going to sit there, unfortunately.

I've a private "fork" of this with some modifications that helped me in some scenarios, but not sure if there's much interest in making it public or even for the community to help out in maintaining/moving forward.

momala454:
i thnk people are interested in your mods

brunogaspar:
Alright, i'll look into making it public and how to make it installable through NPM, not much experience with it, just Composer for PHP :)

Niek:
For those who're interested: I decided to make a similar package based on Playwright instead: https://github.com/Niek/playwright-addons. It's still in development, at the moment it includes an ad blocker + some stealth evasions.

itsdarrylnorris:
I have created a simple PR to avoid this particular issue. https://github.com/berstend/puppeteer-extra/pull/240. If anyone can give that PR a quick look and test it out I will appreciated.

@brunogaspar Let me know if there is anything I can help getting your fork public, feel free to shoot me an email.

Hopefully, this https://github.com/berstend/puppeteer-extra/issues/238 gets resolve soon, so we can all contribute to the same project.

berstend:
Hi all, as mentioned in #238 I'm working my way through the outstanding issues and PRs now.

Making `puppeteer-extra` compatible with the latest Puppeteer release is among the highest priorities. I'm being slowed down by the CI testing a bit, as we have quite an elaborate testing matrix currently to ensure backwards compatibility.

As [mentioned here](https://github.com/berstend/puppeteer-extra/pull/240#issuecomment-654325111) it could make sense to simplify this in the future by having the latest `puppeteer-extra` version always support the latest `puppeteer` version (and encouraging version pinning in case older versions are required).

TL;DR: I'm in the process of merging/testing PRs to support Puppeteer 5 asap. 😄

scottweinert:
[...]

PR #240, opened 2020-07-03
Fixing issue with the setMaxListeners

itsdarrylnorris:
### Problem:
New versions of puppeteer do not longer support the method setMaxListeners, and it makes puppeteer-extra-stealth not usable. 

### Solution 

This PR check if the method available before running it.

-----
Related Issue: https://github.com/berstend/puppeteer-extra/issues/228

### DIFF
--- a/packages/puppeteer-extra-plugin-stealth/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/index.js
@@ -151,8 +151,10 @@ class StealthPlugin extends PuppeteerExtraPlugin {
   }
 
   async onBrowser(browser) {
-    // Increase event emitter listeners to prevent MaxListenersExceededWarning
-    browser.setMaxListeners(30)
+    if (browser && browser.setMaxListeners) {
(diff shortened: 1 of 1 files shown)


Niek:
I think I'd make more sense to support only the latest stable Puppeteer version. So bump dependency to Puppeteer 5 and just remove the line.

berstend:
> I think I'd make more sense to support only the latest stable Puppeteer version. So bump dependency to Puppeteer 5 and just remove the line.

Backwards compatibility has been a feature so far, but instead of maintaining the code to support older versions we can also just list `puppeteer-extra` versions which are compatible with older Puppeteer releases in the readme (if need be).

The latest `puppeteer-extra` version can then be optimized solely for the latest stable Puppeteer (or Playwright) version.

This would definitely make the CI testing setup leaner (currently we have quite the testing matrix).

berstend:
Merging this workaround for now until we've decided how to handle backwards compat in the future. @itsdarrylnorris thanks!

PR #245, opened 2020-07-09
chore(ci): Add GitHub Action based testing

moltar:


.github/workflows/test.yml
```diff
@@ -0,0 +1,69 @@
...
+  test:
+    name: Test Node v${{ matrix.node }} on ${{ matrix.os }} and ${{ matrix.puppeteer_version }}
+
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        puppeteer_version:
+          # - 5.0.0
+          - 2.1.1 # Chromium 79.0.3942.0, Oct 24 2019
```
berstend, line 22: From my perspective testing the two latest major Puppeteer releases should be fine (currently: 5.0.0 and 2.1.1).

Historically I've added more and more versions as they appeared (being somewhat proud of our backwards compatibility) but that made waiting for the CI results quite annoying. 😄

.github/workflows/test.yml
```diff
@@ -0,0 +1,69 @@
...
+          # - 2.0.0 # Chromium 79.0.3942.0, Oct 24 2019
+          # - 1.20.0 # Chromium 78.0.3882.0, Sep 13 2019
+          # - 1.15.0 # Chromium 75.0.3765.0, Apr 26 2019
+          # - 1.9.0 # Chromium 71.0.3563.0, Oct 4, 2018
+          # - 1.6.2 # Chromium 69.0.3494.0, Aug 1, 2018
+        node:
+          - 10
```
berstend, line 29: I'd suggest we test using the latest NodeJS LTS & Current release from now on (v12 & v14, maybe v10 as well if GH actions is fast):
https://nodejs.org/en/about/releases/

.github/workflows/test.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup Node
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node }}
+
+      - name: 'Fix for: error fsevents@2.1.2: The platform "linux" is incompatible with this module.'
+        run: npx json -I -f package.json -e 'this.resolutions={}'
```
berstend, line 45: Haven't tested it yet but we might not need this hack anymore (#164) by using `yarn --ignore-optional` :-)

moltar: Just tried it. Still got an error:

```
error fsevents@2.1.2: The platform "linux" is incompatible with this module.
error Found incompatible module.
```

.github/workflows/test.yml
moltar: Yeah, I'll enabled the 5 back when I am done fixing build / test steps. I just didn't want to run the entire matrix for fixing Windows-related issues only. I just commented out items in the matrix to only run one Windows test.

Actions do run in parallel, but there does seem to be some limit as to how many. Might be dependent on availability of boxes too. We can test different number of items to see how long it takes.

moltar: Full matrix run now (without Windows):

https://github.com/berstend/puppeteer-extra/actions/runs/164138527

.github/workflows/test.yml
moltar: Added Node 10, 12, 14: https://github.com/berstend/puppeteer-extra/actions/runs/164138527

moltar:
Resolved yarn.lock conflict and a did a force push

.github/workflows/test.yml
```diff
@@ -0,0 +1,76 @@
+name: Test
+
+on:
+  - push
+  - pull_request
```
berstend, line 5: Would it make sense to be more specific here so not too many test runs are triggered?

e.g. I've seen this:

```yaml
on:
  pull_request:
    types: [opened, synchronize, reopened]
```


here:
https://github.com/formik/tsdx/blob/master/.github/workflows/nodejs.yml

berstend: They also use https://github.com/bahmutov/npm-install which claims to use the yarn cache better/result in faster tests (haven't tested if that's true).

packages/puppeteer-extra-plugin-devtools/package.json
```diff
@@ -7,10 +7,10 @@
   "author": "berstend",
   "license": "MIT",
   "scripts": {
-    "docs": "exit 0",
-    "test": "eslint --ext .js .",
-    "test-old": "ava -v && eslint --ext .js .",
-    "test-ci-old": "yarn test"
+    "docs": "node -e 0",
+    "lint": "eslint --ext .js .",
+    "test": "run-s lint",
```
berstend, line 12: Did we accidentally drop `ava` code testing here? I don't mind splitting `lint` into it's own script, could still make sense to run it as well when `test` is called

.github/workflows/test.yml
moltar: > Would it make sense to be more specific here so not too many test runs are triggered?

Done!

> They also use https://github.com/bahmutov/npm-install which claims to use the yarn cache better/result in faster tests (haven't tested if that's true).

I do like to use cache too. But travis config had explicit anti-cache setting:

```ts
# Prevent potential issues
cache:
  npm: false
  yarn: false
```

If you think it is ok to enable cache, then we can.

packages/puppeteer-extra-plugin-devtools/package.json
moltar: Oops, got eye! I put it back.

.github/workflows/test.yml
berstend: We should definitely try it out, the Travis config is a bit messy as I had issues testing out different puppeteer versions 😄

moltar: Done! Added bahmutov/npm-install

berstend: It could make sense to add `yarn list puppeteer` somewhere so we can verify that using a different puppeteer version actually works as intended. :-)

moltar: There is:

```yaml
      - name: debug
        run: |
          yarn list --pattern "puppeteer|puppeteer-extra"
```

moltar: I had to change it from plain `yarn list` because that is deprecated and throws warnings.

moltar: Sample output:

```
Run yarn list --pattern "puppeteer|puppeteer-extra"
yarn list v1.22.4
├─ @cliqz/adblocker-puppeteer@1.9.2
├─ @types/puppeteer@2.0.1
├─ puppeteer-extra-plugin-adblocker@2.11.5
├─ puppeteer-extra-plugin-anonymize-ua@2.2.11
├─ puppeteer-extra-plugin-block-resources@2.2.6
├─ puppeteer-extra-plugin-click-and-wait@2.2.6
├─ puppeteer-extra-plugin-devtools@2.2.11
├─ puppeteer-extra-plugin-flash@2.2.7
├─ puppeteer-extra-plugin-font-size@2.2.6
├─ puppeteer-extra-plugin-recaptcha@3.1.13
├─ puppeteer-extra-plugin-repl@2.2.7
├─ puppeteer-extra-plugin-stealth@2.4.13
├─ puppeteer-extra-plugin-user-data-dir@2.2.6
├─ puppeteer-extra-plugin-user-preferences@2.2.6
├─ puppeteer-extra-plugin@3.1.6
├─ puppeteer-extra@3.1.12
├─ puppeteer-firefox@0.5.1
└─ puppeteer@5.0.0
Done in 2.23s.
node_modules/puppeteer-extra/dist/index.cjs.js: ASCII text, with CRLF, LF line terminators
```

moltar:
Weird, the tests seem to be stuck after:

```
puppeteer-extra-plugin-devtools:   ✔ lib › RemoteDevTools › will throw when missing webSocketDebuggerUrl
```

Both in GH and Travis.

I don't think I changed anything in the tests since the last successful run.

Could it be commit #562a0ea? Where I added a missing test back?

moltar:
Right, the package name for which the test was added matches. Hmmm... Going to look into it locally.

berstend:
> Could it be commit #562a0ea? Where I added a missing test back?

Yeah, super weird.

I'd say we leave the `devtools` test out for now, could well be that the test isn't written in a robust way (the plugin isn't really popular so it's fine).

moltar:
Ok, done. Commented out the test.

Shall we keep Travis for now or shall we remove it?

berstend:
> Shall we keep Travis for now or shall we remove it?

Let's run them in parallel until we feel comfy with the new setup. :-)

.github/workflows/test.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: debug
+        run: |
+          yarn list --pattern "puppeteer|puppeteer-extra"
+          file node_modules/puppeteer-extra/dist/index.cjs.js
+
+      - name: test
+        run: yarn test-ci
+
+      - name: test with xvfb
```
berstend, line 74: Oh, I think we're effectively running the tests twice now? It should be sufficient to only once run the tests with xvfb support (in case there are headful tests)

moltar: Yeah, I am not too familiar with this. I figured one way runs them headless and the other will run them headed?

Or is it not the case?

moltar: Do we want to test in headed mode only then? (with xvfb on)

moltar:
`npm-install` doesn't seem to be that reliable

moltar:
Actually it seems to be a network error:

> info There appears to be trouble with your network connection. Retrying...

.github/workflows/test.yml
berstend: Do we want to test in headed mode only then? (with xvfb on)

Yup

moltar: Ok, done.

But you don't think there might be some differences between headed and headless mode? There are no detections that check for headless?

Sesamestrong:
May I ask why this uses `rimraf` instead of `rm -rf`? I've never used it before and would like to know why it is preferred; the general idea I got from the package description was that it was just a node version of `rm -rf`.

berstend:
> May I ask why this uses `rimraf` instead of `rm -rf`? I've never used it before and would like to know why it is preferred; the general idea I got from the package description was that it was just a node version of `rm -rf`.

`rm` doesn't exist on windows :-) Also rimraf fails more gracefully IIRC (if e.g. the folder doesn't exist).

.github/workflows/test.yml
berstend: > But you don't think there might be some differences between headed and headless mode? There are no detections that check for headless?

Not that I recall. :-) Headless tests will run normally even in the presence of a virtual frame buffer which is needed for headful tests.

moltar:
> May I ask why this uses `rimraf` instead of `rm -rf`? I've never used it before and would like to know why it is preferred; the general idea I got from the package description was that it was just a node version of `rm -rf`.

Just to confirm what @berstend said:

Yes, `rm -rf` only works on Unix-like systems, while Windows users do not have that command.

One of the things I wanted achieve with this PR is to get Windows tests to work, and therefore enable the development experience for Windows users.

That meant getting rid of Unix-specific commands and constructs such as `rm`, and `&&` for command joining.

moltar:
Do we need anything else before we can merge this?

packages/puppeteer-extra-plugin-recaptcha/package.json
```diff
@@ -13,29 +13,24 @@
...
+    "clean": "rimraf dist/*",
     "tscheck": "tsc --pretty --noEmit",
-    "build": "yarn clean && tsc --project tsconfig.json --module commonjs && rollup -c rollup.config.ts && yarn ambient-dts",
-    "docs": "exit 0",
-    "docs2": "rm -rf docs/* && typedoc --module commonjs --readme none --target ES6 --theme markdown --excludeNotExported --excludeExternals --excludePrivate --out docs --mode file src/index.ts",
-    "test": "TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' ava -v src/**.test.ts",
-    "test-ci": "echo pwd && yarn build && TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' ava --fail-fast -v dist/*.test.js",
-    "ambient-dts": "yarn ambient-dts-copy && yarn ambient-dts-fix-path",
-    "ambient-dts-copy": "copyfiles -u 1 \"src/**/*.d.ts\" dist",
-    "ambient-dts-fix-path": "replace-in-files --string='/// <reference path=\"../src/' --replacement='/// <reference path=\"../dist/' 'dist/**/*.d.ts'"
```
berstend, line 25: Did we accidentally lose `ambient-dts-fix-path` here? It's unfortunately needed to make TypeScript pick up our ambient declarations

moltar: I removed that, because there were no files that ever matched that regex..

berstend: > I removed that, because there were no files that ever matched that regex..

You mean `dist/**/*.d.ts`? This will only exist/be populated after the build step (ts to js). :-)

berstend: The `.d.ts` files that will be copied to `dist/` and need to be patched can be seen here: https://github.com/berstend/puppeteer-extra/tree/master/packages/puppeteer-extra-plugin-recaptcha/src

moltar: You are right. I did check to see if there were any `reference path` references in `.d.ts` files, but didn't see any, because I was relying on a human sight :)

Then I used `grep` and found that indeed there is a reference.

Fixed!

berstend:
Alright, the Travis test passed as well. I'm gonna merge this in now (no point in having others review this as well) and we're gonna run this in parallel with Travis for a while (more testing doesn't hurt haha).

Other GH actions planned for the future:
- Run `yarn docs` and push the changes back to the PR (ideally only for packages that have changed source files)
- Decide which release process to follow (conventional/semantic commits) and how publishing/version bumping is being done (changesets, run `lerna version` locally, automatically derived from semantic commit messages, etc)
- Implement auto-publishing based on the above (would be nice to publish to GH as well as NPM)

Thanks @moltar 👍

moltar:
> Alright, the Travis test passed as well. I'm gonna merge this in now (no point in having others review this as well) and we're gonna run this in parallel with Travis for a while (more testing doesn't hurt haha).
> 
> Other GH actions planned for the future:
> 
>     * Run `yarn docs` and push the changes back to the PR (ideally only for packages that have changed source files)
> 
>     * Decide which release process to follow (conventional/semantic commits) and how publishing/version bumping is being done (changesets, run `lerna version` locally, automatically derived from semantic commit messages, etc)
> 
>     * Implement auto-publishing based on the above (would be nice to publish to GH as well as NPM)
> 
> 
> Thanks @moltar 👍

Maybe create separate issues for these?

I can work on these. I've used Semantic Release extensively, but never with lerna. I've seen that there is a plugin available for that, but need to experiment.

berstend:
> Maybe create separate issues for these?

We already have so many open issues 🙈 It's an eternal battle against windmills I feel 😄 

> I can work on these. I've used Semantic Release extensively, but never with lerna. I've seen that there is a plugin available for that, but need to experiment.

The main question is how publishing is triggered, I've checked out other lerna monorepos and some use a Tag based trigger (where tags need to be pushed locally through the lerna cli).

An approach I favor the most would be to use conventional commits + semantic release and derive the version numbers automatically + publish afterwards. Icing on the cake would be Github releases with an auto-generated Changelog based on the conventional commits.

There's a bunch of different repos trying to tackle semantic release + monorepos, e.g. https://github.com/dhoulb/multi-semantic-release (haven't checked them out in more detail yet).

If you want to give it a shot (maybe we can create a single umbrella issue for them 😅 ) feel free to do so. :-)

PR #259, opened 2020-07-17
Added puppeteer-extra-plugin-minmax to README

Stillerman:


brunogaspar:
Not sure if we should add a list of plugins from the community to separate from the official ones.

What you think @berstend ?

berstend:
Hmm, I favor adding the plugins under a community headline, so it's clear they're not under the control of the main project. :-)

But nice to see new plugins being developed, thanks @Stillerman

I have some time tomorrow and will update the PR with a respective change.

moltar:
Agree with @berstend we should have a separate section with the community ones.

berstend:
Uhm, given how cumbersome it still is to modify PRs from other people I'm gonna merge this in (so Stillerman gets attribution) and modify the readme afterwards. :-)

berstend:
Here's the commit with the updated Community Plugin section: https://github.com/berstend/puppeteer-extra/commit/32e2eb442af4ed1df3768d462638371fffecb80c

PR #268, opened 2020-07-29
Fixing problem with NCC

itsdarrylnorris:
### Problem:

I am having an issue compiling a project using puppeteer-extra-plugin-stealth with NCC.

After some investigation, it looks like the problem it's related to a sub-dependency of the package merge-deep. I created a PR to fix this particular problem in the merge-deep https://github.com/jonschlinkert/merge-deep/pull/14. However, that project does not look like it has been active for the last two years, and I wonder if my PR will even get merge. 

Here is the NCC issue explaining the problem: https://github.com/vercel/ncc/issues/500

### DIFF
--- a/packages/puppeteer-extra-plugin-recaptcha/package.json
+++ b/packages/puppeteer-extra-plugin-recaptcha/package.json
@@ -64,7 +64,6 @@
   },
   "dependencies": {
     "debug": "^4.1.1",
-    "deepmerge": "^4.2.2",
     "puppeteer-extra-plugin": "^3.1.7"
   },
   "peerDependencies": {
--- a/packages/puppeteer-extra-plugin-recaptcha/package.json
+++ b/packages/puppeteer-extra-plugin-recaptcha/package.json
@@ -64,7 +64,7 @@
   },
   "dependencies": {
     "debug": "^4.1.1",
-    "merge-deep": "^3.0.2",
+    "deepmerge": "^4.2.2",
     "puppeteer-extra-plugin": "^3.1.7"
   },
   "peerDependencies": {
--- a/packages/puppeteer-extra-plugin/package.json
+++ b/packages/puppeteer-extra-plugin/package.json
@@ -64,7 +64,7 @@
   "dependencies": {
     "@types/debug": "^4.1.0",
     "debug": "^4.1.1",
-    "merge-deep": "^3.0.1"
+    "deepmerge": "^4.2.2"
   },
   "peerDependencies": {
     "puppeteer-extra": "*"
--- a/packages/puppeteer-extra-plugin/src/index.ts
+++ b/packages/puppeteer-extra-plugin/src/index.ts
@@ -2,7 +2,7 @@ import debug, { Debugger } from 'debug'
 import * as Puppeteer from 'puppeteer'
 
 /** @private */
-const merge = require('merge-deep')
+const merge = require('deepmerge')
 
 export interface PluginOptions {
   [key: string]: any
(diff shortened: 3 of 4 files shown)


packages/puppeteer-extra-plugin-recaptcha/package.json
```diff
@@ -64,7 +64,7 @@
   },
   "dependencies": {
     "debug": "^4.1.1",
-    "merge-deep": "^3.0.2",
+    "deepmerge": "^4.2.2",
```
jozsi, line 67: Is this dependency needed at all for the recaptcha plugin? I don't see any reference to it.

packages/puppeteer-extra-plugin/package.json
```diff
@@ -64,7 +64,7 @@
   "dependencies": {
     "@types/debug": "^4.1.0",
     "debug": "^4.1.1",
-    "merge-deep": "^3.0.1"
+    "deepmerge": "^4.2.2"
```
jozsi, line 67: Same quetsion as above.

packages/puppeteer-extra-plugin-recaptcha/package.json
brunogaspar: It's being used on line 77 to merge the passed opts with the defaults

brunogaspar: As @jozsi pointed out, this is not being used on the recaptcha plugin, therefore the dependency can most likely be removed.

packages/puppeteer-extra-plugin/package.json
itsdarrylnorris: @jozsi We are using it here https://github.com/berstend/puppeteer-extra/blob/d5d82b4506a6ca514b7852709f30a7815f8a1304/packages/puppeteer-extra-plugin/src/index.ts#L5

itsdarrylnorris:
After doing some testing with deepmerge, and puppeteer-extra I found out an issue with deepmerge that it's probably a blocker to use deepmerge, because it does not work with Set and we are using set here https://github.com/berstend/puppeteer-extra/blob/40c47f158170b7a10f091a4acdd235eb4602c0c6/packages/puppeteer-extra-plugin-stealth/index.js#L82

I created an issue in deepmerge about this: https://github.com/TehShrike/deepmerge/issues/202

itsdarrylnorris:
I tried to use [object-merge-advanced](https://www.npmjs.com/package/object-merge-advanced), but I ran into the same issue as merge-deep.

Any thoughts on how to move forward?

One possible alternative is to use [this fork](https://github.com/itsdarrylnorris/merge-deep/tree/updating/clone-deep), which I made to merge-deep and it fixes the issue with NCC.

Thoughts?

berstend:
What's the latest on this, is there a drop in replacement that works with Sets a well?
I'm creating a new base plugin (to support the new Playwright + Puppeteer rewrite) so now would be a good time to change that stuff. 😄

berstend:
The author responded and mentioned using `isPlainObject`, I did some quick testing on runkit and think this is fixed?

image

itsdarrylnorris:
@berstend I think using deepmerge  with `isPlainObject` it's the way to go. If I am not wrong I ran into other issue with it, but this is unrelated to puppeteer-extra.

Do you want me to make update to this PR to using deepmerge  with `isPlainObject`? Or are you going to include this as part new plugin re-write?

berstend:
@itsdarrylnorris apologies for the "slight" delay on this 😄 This will be fixed in #303, with [this commit](https://github.com/berstend/puppeteer-extra/pull/303/commits/48c3eae5ecf57468738ea9adedb5e4db5a5432e2).

itsdarrylnorris:
@berstend No worries, I stopped using NCC it causes too many issues. Thanks for the #303 hard work.

PR #272, opened 2020-08-02
Support script tag detection for different domains

verglor:
Some sites use `https://www.recaptcha.net/recaptcha/api.js` instead of `https://www.google.com/recaptcha/api.js` (e.g. [Mintos](https://www.mintos.com/en/login)).


berstend:
Published as `puppeteer-extra-plugin-recaptcha@3.1.15`

Issue #207, opened 2020-05-26
Does puppeteer-extra use js to replace browser settings?


qo4on:
If so stealth is no longer a stealth. Read this [article](https://cpa.rip/stati/antidetect-palivo/) (Google translate to English).

StevenVeshkini:
Yeah, seems like Headless Chrome can be detected quite easily with the techniques in that article...For example, running:

`if( Object.getOwnPropertyNames(navigator)[0] ) alert('fake parameters detected');`

after loading a page in the stealth plugin will cause the alert, but wont in regular Chrome...

qo4on:
It's very intresting to know what @berstend and @Niek think about it.

Niek:
That's easy to work around: always copy the prototype of an object, make changes and then copy it back.
```javascript
const newProto = navigator.__proto__;
Object.defineProperty(newProto, 'languages', {get: () => ['my', 'non-existing', 'languages']});
navigator.__proto__ = newProto;
navigator.languages
     (3) ["my", "non-existing", "languages"]
Object.getOwnPropertyNames(navigator)
     []
```

qo4on:
Puppeteer stealth does not work. Try to create a gmail account and you get an error.

berstend:
Fixed in #275, #276, #277, #278 😄

Issue #239, opened 2020-07-01
Error stack trace inconsistency


Sesamestrong:
These are a few live methods of detection which currently work (and are in use by bot detection services) to distinguish `puppeteer-extra-plugin-stealth` running headless from normal Chrome.
I think I know a fix for one or two of these, but I think that the rest of them are up to other, more V8-savvy maintainers to evade. I will post more detection methods in the coming days.
1. `document.createElement`
```js
try{
  document.createElement("dummy value")
} catch({stack}) {
  if(stack.split("\n")[1].includes("Object.apply (<anonymous>"))
    console.log("This is puppeteer");
}
```
2. `window` dimensions
```js
if(window.outerHeight-window.innerHeight>160&&window.outerWidth-window.innerHeight>160)
  console.log("This is puppeteer");
```
3. New detection method of `navigator.webdriver`
```js
if(!!Object.getOwnPropertyDescriptor(navigator.__proto__,"webdriver"))
  console.log("This is puppeteer");
```
Fix: I am not sure about this at all, but possibly
```js
delete navigator.__proto__.webdriver;
```
4. New detection method of `navigator.languages`
```js
if(!!Object.getOwnPropertyDescriptor(navigator, "languages"))
  console.log("This is puppeteer");
```
5. `console.debug`
```js
if((console.debug+"").includes("return"))
  console.log("This is puppeteer");
```

Bllacky:
Not sure if this is correct, but I tried this versions of removing navigator:
`const newProto = Object.getPrototypeOf(navigator);
delete Object.getPrototypeOf(navigator).webdriver;
Object.setPrototypeOf(navigator,newProto);`

But I think I might be doing something wrong.

JimmyLaurent:
Another ones used by cloudflare:
```js
Array.isArray(navigator.plugins); // true with stealth hacks otherwise false
Array.isArray(navigator.mimeTypes); // true with stealth hacks otherwise false
```

brunogaspar:
@Sesamestrong Thanks for the tricks, here's some feedback:

> 1. `document.createElement`

Yields false to me.

> 2. `window` dimensions

This behaves the other way around for me hehe, in one of my bots, this actually yields false, but on the real browser, this yields true, doesn't that accurate.

> 3. New detection method of `navigator.webdriver`

I have this applied on my end aswell, but sometimes this doesn't work, haven't digged much into it.

> 4. New detection method of `navigator.languages`

Yields false to me.

[...]

PR #275, opened 2020-08-06
fix(plugin-stealth): Harden navigator.languages evasion

berstend:
The way we modified `navigator.languages` was from back in the old days and could be detected in various ways (https://github.com/berstend/puppeteer-extra/issues/239). 😅 

This PR changes our code to use the prototype instead and adds two leak tests that would otherwise fail.

edit, there were more occurrences: #276, #277, #278

Fixes: #207 (https://cpa.rip/stati/antidetect-palivo/)


berstend:

Issue #132, opened 2020-01-07
Locale parameter with qvalues in user-agent-override evasion


silentvick:
Maybe it's just a documentation issue, but right now for user-agent-override it says that we can pass a locale string with qvalues in it. It seems though that it leads to broken values for navigator.languages and Accept-Language header. Like this:

```javascript
const UserAgentOverride = require("puppeteer-extra-plugin-stealth/evasions/user-agent-override");

const ua = UserAgentOverride({
  userAgent: 'Teapot/10.1',
  locale: 'ru-RU,ru;q=0.9,en-US;q=0.8,en;q=0.7',
});

puppeteer.use(ua);

// navigator.languages => [ 'ru-RU', 'ru;q=0.9', 'en-US;q=0.8', 'en;q=0.7' ]
```

Notice the duplication:
```
Accept-Language: ru-RU,ru;q=0.9;q=0.9,en-US;q=0.8;q=0.8,en;q=0.7;q=0.7
```

Everything's fine if you pass it without qvalues:

```javascript
const UserAgentOverride = require("puppeteer-extra-plugin-stealth/evasions/user-agent-override");

const ua = UserAgentOverride({
  userAgent: 'Teapot/10.1',
  locale: 'ru-RU,ru,en-US,en',
});

puppeteer.use(ua);

// navigator.languages => [ 'ru-RU', 'ru', 'en-US', 'en' ]
// Accept-Language: ru-RU,ru;q=0.9,en-US;q=0.8,en;q=0.7
```

berstend:
Interesting. The `user-agent-override` evasion will just pass on the `locale` string:

```js
page._client.send('Network.setUserAgentOverride', { acceptLanguage: locale }})
```

The [CDP documentation](https://chromedevtools.github.io/devtools-protocol/tot/Network#method-setUserAgentOverride) doesn't say much (incl. typo 😄) regarding possible values:
> string - Browser langugage to emulate.

Are you setting the locale differently as well? (e.g. a `--lang` launch arg)

silentvick:
No, only through the user-agent-override. It's easy to reproduce even with vanilla puppeteer:

```javascript
const puppeteer = require('puppeteer');

(async () => {
    const browser = await puppeteer.launch({
        headless: true,
        ignoreHTTPSErrors: true,
        args: [
            '--no-sandbox',
            '--disable-setuid-sandbox',
        ]
    });

    const page = await browser.newPage();

[...]

PR #276, opened 2020-08-06
fix(plugin-stealth): Harden navigator.vendor evasion

berstend:
- While improving docs I found another bad occurrence of overriding native APIs. Added fix + regression tests.
- Add missing readmes for `navigator.vendor` and `user-agent-override`.
- Also fixes https://github.com/berstend/puppeteer-extra/issues/132 by clarifying `qvalues` & `user-agent-override`.


berstend:

PR #278, opened 2020-08-06
fix(plugin-stealth): Harden navigator.plugins evasion

berstend:
Alright, this is the last one of those 😅 

With that our `navigator` props are peachy clean and this will return an empty array when using stealth:

```js
Object.getOwnPropertyNames(navigator)
// ==> []
```


berstend:

PR #279, opened 2020-08-06
fix(plugin-stealth): Remove console.debug evasion (obsolete)

berstend:
I have no idea why we had this, copy pasta from the old days.

Even the test check for different behavior in vanilla/stealth? I'm confused.

I checked `console.debug` in headless and it's seems perfectly normal to me, whereas our patch does more harm than good.


berstend:
Tests are failing due to yarn outage 🤦

berstend:

PR #291, opened 2020-08-15
feat(plugin-stealth): Add shared stealth utils, harden and rewrite some evasions

berstend:
This PR:
- adds new, hardened & documented utility functions to patch native browser APIs without leaving traces
- adds a truck load of tests against these utility functions
- updates `webgl.vendor` & `media.codecs` evasions to use the new utility functions
- rewrites `navigator.permissions` - the code was extremely ugly and I couldn't reproduce the issue initially

Thanks to @Sesamestrong for initiating this and finding weird edge-cases to test against 👍 

### Utility functions

We had some lingering more fundamental issues with our approach to patch native browser APIs. JS Proxies are definitely the way to go but they had two known issues: A couple of `toString` related quirks and they would reveal their presence in stack traces.

Both these issues are now fixed in a pretty robust way. I'm currently not aware of any other Proxy-usage related leaks. 😄💪 

---

### `navigator.permissions`
image

I'm unable to reproduce the issue and assume it's been fixed since then in headless:
image


Niek:
Very impressive work! 👍🏻

berstend:
According to travis the permissions thing might actually still be an issue (or just a synthetic test failure):
```
The following fingerprints failed: [ { name: 'HEADCHR_PERMISSIONS', consistent: 1, data: {} } ]
```

berstend:
This makes no sense, fpscanner is checking specifically that the permission values are inconsistent? I think this might be a bug on their end: https://github.com/antoinevastel/fpscanner/blob/b2594d4508839c8be2554d9d090a6027c257d8d8/test/test.js#L73

berstend:
Oh well, I just rewrote the `navigator.permissions` in a nice way to make the tests happy.

berstend:
This feels pretty solid and well tested now, merging the PR so we can work on fixing problems with the new utility functions.

berstend:
Published in `puppeteer-extra-plugin-stealth@2.6.0`

PR #293, opened 2020-08-18
Add package: extract-stealth-evasions

Niek:
This is a simple script to extract the stealth evasions to a minified JS file that can be used in browsers/non-puppeteer environments.


berstend:
The wildcard made CI happy 😄

PR #294, opened 2020-08-18
chore(extract-stealth-evasions): Add bin to package.json

berstend:
This should make it possible to use e.g. `npx extract-stealth-evasions` without installing anything. I also wildcarded the other relevant deps just in case and VSCode did some linting 😄


berstend:
> Awesome :) Can you maybe update the readme as well with the npx command?

I'm not 100% yet that npx will work after publishing this change 😄 once that's confirmed I'll update the readme :)

berstend:
There's another issue with the `bin` stuff which I'll fix

PR #295, opened 2020-08-18
Changes to the extract package

Niek:
A minor change (typo fix) and a new feature: rebuild a checked-in stealth.min.js upon pushes to the evasions files.

Point of discussion: maybe we should push to a separate branch (that we can serve e.g. in a GitHub page or CDN)?


.github/workflows/extract-stealth.yml
```diff
@@ -0,0 +1,30 @@
...
+        run: |
+          cd packages/extract-stealth-evasions
+          npm install
+          node index.js
+
+      - name: Commit stealth.min.js
+        uses: EndBug/add-and-commit@v4
+        with:
+          add: 'packages/extract-stealth-evasions/stealth.min.js'
+          force: true
```
Niek, line 27: As commented, it might be better to push to a separate branch (by adding `ref: 'stealth-js'` for example).

berstend: Yeah, I think using a dedicated branch for this is best as it gives us more flexibility later on (also reduces noise on `master`). :-)

berstend: If we're lazy we can just https://gitcdn.xyz/ for the CDN part. We could also make a small snippet to load the stuff into devtools based on the CDN url (https://stackoverflow.com/questions/7474354/include-jquery-in-the-javascript-console)

berstend (commented):
Looks good, I think it makes sense switching to a designated branch for the compiled output before merging this in. :-)

packages/extract-stealth-evasions/index.js
```diff
@@ -57,7 +57,8 @@ puppeteer
 
     fs.writeFile(
       file,
-      (await minify(scripts, { toplevel: true })).code,
+      '/* Auto-generated... do not update manually */\n' +
```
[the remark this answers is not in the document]
berstend: And maybe a link to the repo, in case someone finds this file without context 😄

berstend (changes requested):
As mentioned in comments

brunogaspar (approved):

berstend (approved):

PR #298, opened 2020-08-21
Add non minified file

momala454:


Niek:
Created a better PR in https://github.com/berstend/puppeteer-extra/pull/308

PR #304, opened 2020-08-31
feat(plugin-recaptcha): Improve invisible iframe support

### LINKED ISSUE #300: Support for Invisible reCaptcha V2 in Puppeteer Extra Plugin Recaptcha
Is anyone having success with solving Invisible reCaptchas with the puppeteer-extra-plugin-recaptcha? I see that 2captcha has described the process needed for them to solve the invisible recaptchas: https://2captcha.com/2captcha-api#invisible

If no one has has tried to solve this issue, I'd be happy to take a crack at it based on the documentation by 2captcha. I just wanted to find out if it had been tried and learn from any roadblocks encountered.

bclougherty:
This PR fixes #300 

Captchas with 'invisible' in the src can be shown via js, at which point the plugin fails to find them.

My patch uses something equivalent to jquery's `is(':visible')` to check whether the captcha is visible to the user, rather than relying on the src value.

### DIFF
--- a/packages/puppeteer-extra-plugin-recaptcha/src/content.ts
+++ b/packages/puppeteer-extra-plugin-recaptcha/src/content.ts
@@ -29,7 +29,7 @@ export class RecaptchaContentScript {
     props.reduce((a, e) => ({ ...a, [e]: o[e] }), {})
 
   // make sure the element is visible - this is equivalent to jquery's is(':visible')
-  private _isVisible = (elem: any) => !!( elem.offsetWidth || elem.offsetHeight || (typeof elem.getClientRects === 'function' && elem.getClientRects().length )))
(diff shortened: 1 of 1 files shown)


berstend (changes requested):
@bclougherty sweet!

there's what looks like a linting issue according to the test runner, could you fix that? :-)

```bash
src/content.ts:32:161 - error TS1005: ';' expected.

32   private _isVisible = (elem: any) => !!( elem.offsetWidth || elem.offsetHeight || (typeof elem.getClientRects === 'function' && elem.getClientRects().length )))
                                                                                                                                                                   ~


Found 1 error.
```

bclougherty:
@berstend Sorry about that, had a couple stray characters in there. It should be good now.

Niek (approved):

berstend (approved):

berstend:
Published as `puppeteer-extra-plugin-recaptcha@3.1.18` - thanks @bclougherty 👍

PR #307, opened 2020-09-05
fake cpu cores to custom number

gyaaniguy:
Hello 
The idea is to specify the number of cpu cores reported by navigator.hardwareConcurrency. I used this option to make my bot appear to have different profiles. 

    puppeteer.use(require('puppeteer-extra-plugin-stealth/evasions/navigator.hardwareConcurrency')({'cores':4}));

I am not sure if this is 'okay' . Am always a lil nervous about PR's . Please let me know if I am doing something silly.


packages/puppeteer-extra-plugin-stealth/evasions/navigator.hardwareConcurrency/index.js
```diff
@@ -0,0 +1,32 @@
...
+    get name() {
+        return 'stealth/evasions/navigator.hardwareConcurrency'
+    }
+
+
+    async onPageCreated(page) {
+    let cores = this.opts.cores || false;
+        await page.evaluateOnNewDocument((cores) => {
+            let originalCores = navigator.hardwareConcurrency;
+             cores = cores || originalCores;
```
Niek, line 23: This line seems to be doing nothing, since you're doing the same in the assignment below

packages/puppeteer-extra-plugin-stealth/evasions/navigator.hardwareConcurrency/index.js
```diff
@@ -0,0 +1,32 @@
...
+            delete Object.getPrototypeOf(navigator).hardwareConcurrency;
```
Niek, line 24: Is there a specific reason to delete and then set? I think only setting should be sufficient.

Niek (changes requested):
Functionality is looking good, I added some comments on the code.

packages/puppeteer-extra-plugin-stealth/evasions/navigator.hardwareConcurrency/index.js
gyaaniguy: yeah it seemed hardwareConcurreny is readonly. I could not reset its value.
Its important to note, this changes the order of properties of navigator. Something that was being flagged in an amIunique website

packages/puppeteer-extra-plugin-stealth/evasions/navigator.hardwareConcurrency/index.js
gyaaniguy: Indeed. An oversight, should I add a commit ?

Niek: Sure 👍🏻

packages/puppeteer-extra-plugin-stealth/evasions/navigator.hardwareConcurrency/index.js
Niek: Can you try this instead?
```js
Object.defineProperty(Object.getPrototypeOf(navigator), 'hardwareConcurrency', { get: () => cores || originalCores });
```
That should do the job

Niek: Can you try to overwrite is @gyaaniguy? That should resolve the order of properties issue.

gyaaniguy: yeah sure. Sorry I got pre-occupied. This project didn't come to fruition.

gyaaniguy:
Made the changes! Although I messed up my commits. Checked as well and it seems to be okay

prescience-data:
It would be a good idea if `.idea/` was added to `.gitignore` I think, I always forget to remove that too 🤣

brunogaspar:
> It would be a good idea if `.idea/` was added to `.gitignore` I think, I always forget to remove that too 🤣

That's a bad practice, since `.idea` or any other IDE specific files does not belong to the project.

Just add `.idea` to your global `.gitignore` file (there's plenty of articles/tutorials on how to do that).

For the `package-lock.json`, it really depends, but i think that for packages it's not entirely needed and that file can be added to the project `.gitignore`.

prescience-data:
> > It would be a good idea if `.idea/` was added to `.gitignore` I think, I always forget to remove that too 🤣
> 
> That's a bad practice, since `.idea` or any other IDE specific files does not belong to the project.
> 
> Just add `.idea` to your global `.gitignore` file (there's plenty of articles/tutorials on how to do that).
> 
> For the `package-lock.json`, it really depends, but i think that for packages it's not entirely needed and that file can be added to the project `.gitignore`.

Interesting, I hadn't thought about it being bad practice in regards to defining concerns etc.

Personally I do have mine in global `.gitignore`, I was referring more to problems I ran into with other open source projects I've worked on in the past, where people that didn't have it added globally were merging in `.idea`. So it became easier to ".idea-proof" it by adding it to the local `.gitignore` and I just kept doing that ever since. 

Thinking about it now though, you are right - it's really just band-aiding loose quality control on merges, so I think I'll stop doing that.

brunogaspar:
People are just aware of the global gitignore. I've seen a shit load of pull requests on for example Laravel where people want to add that stuff 🤣

Niek (approved):

packages/puppeteer-extra-plugin-stealth/evasions/navigator.hardwareConcurrency/index.js
```diff
@@ -0,0 +1,29 @@
...
+
+const {PuppeteerExtraPlugin} = require('puppeteer-extra-plugin')
+
+/**
+ * Allows faking CPU core number. This allows creating different profiles.
+ * Will overwrite `navigator.hardwareConcurrency` property.
+ */
+class Plugin extends PuppeteerExtraPlugin {
+    constructor(opts = {}) {
+        super(opts)
```
berstend, line 11: There's some linting issues in the file (using 4 spaces instead of 2). Quick way to fix those:

```bash
npx prettier@1.19.1 --write evasions/navigator.hardwareConcurrency/index.js
```

packages/puppeteer-extra-plugin-stealth/evasions/navigator.hardwareConcurrency/index.js
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    get name() {
+        return 'stealth/evasions/navigator.hardwareConcurrency'
+    }
+
+
+    async onPageCreated(page) {
+        await page.evaluateOnNewDocument((cores) => {
+            let originalCores = navigator.hardwareConcurrency;
```
berstend, line 21: - `originalCores` can be `const` as the value is never changing
- I'd feel more comfortable if we do a quick check if `cores` is defined and if not just `return` without any modification (also no need for `originalCores` then)

berstend (changes requested):
I've added some comments to the code. I suggest adding `prettier` to your code editor so the files are automatically formatted correctly when saving them (or just using VSCode).

packages/puppeteer-extra-plugin-stealth/evasions/navigator.hardwareConcurrency/readme.md
```diff
@@ -0,0 +1,24 @@
...
+### class: [Plugin](https://github.com/berstend/puppeteer-extra/blob/e6133619b051febed630ada35241664eba59b9fa/packages/puppeteer-extra-plugin-stealth/evasions/navigator.webdriver/index.js#L9-L23)
+
+- `opts` (optional, default `{}`)
+
+**Extends: PuppeteerExtraPlugin**
+
+Allows faking CPU core number. This allows creating different profiles.
+Will overwrite `navigator.hardwareConcurrency` property.
+
+####Usage:
```
berstend, line 18: It doesn't seem like `Generated by documentation.js.` is true here 😄 Please add the text to the JSDoc comment in the code and run `yarn docs-for-evasions` in the `puppeteer-extra/packages/puppeteer-extra-plugin-stealth`folder to auto-generate the markdown file based on the code comments.

berstend (changes requested):

packages/puppeteer-extra-plugin-stealth/evasions/navigator.hardwareConcurrency/index.js
gyaaniguy: Would `this.opts.cores || navigator.hardwareConcurrency` be okay ? or something like
getattr(this.opts, 'cores', navigator.hardwareConcurrency)

packages/puppeteer-extra-plugin-stealth/evasions/navigator.hardwareConcurrency/readme.md
gyaaniguy: oops. I have truncated the readme. Added comments. but yarn won't run on my windows system.

packages/puppeteer-extra-plugin-stealth/evasions/navigator.hardwareConcurrency/index.js
berstend: What I meant is let's not overwrite the property at all if theres no need. :-)

Example:

```js
async onPageCreated(page) {
  if (!this.opts.cores || !Number.isInteger(this.opts.cores)) {
    return
  }

  await page.evaluateOnNewDocument((cores) => {
    Object.defineProperty(Object.getPrototypeOf(navigator), 'hardwareConcurrency', { get: () => cores });
  }, this.opts.cores)
}
```

packages/puppeteer-extra-plugin-stealth/evasions/navigator.hardwareConcurrency/readme.md
berstend: Would be good if you find a way (WSL or something), I also noticed that tests are missing which are required for evasions (otherwise the code wouldn't be maintainable anymore). :-)

packages/puppeteer-extra-plugin-stealth/evasions/navigator.hardwareConcurrency/index.js
gyaaniguy: Ah. That makes a lot more sense. Made the change.

packages/puppeteer-extra-plugin-stealth/evasions/navigator.hardwareConcurrency/readme.md
gyaaniguy: Thanks a lot for all this. I feel being mentored :)
hmm yarn works on WSL. but fails with 'loop: not found'. I believe its failing at 
`"postdocs-for-evasions": "cd ./evasions && loop \"npx prettier --write readme.md\""`

I am not really great with nodejs. writing test would be hard and could be prone to error ..

berstend:
Closing in favor of #358

Issue #324, opened 2020-09-15
Automation framework detected by pixelscan


Somberor:
hey guys , 
any idea how can pixelscan.net detect my puppeteer even with stealth on?
http://prntscr.com/uhs1aq

This is a sample of the code:
`const puppeteer = require('puppeteer-extra');

    const pluginStealth = require('puppeteer-extra-plugin-stealth');

    puppeteer.use(pluginStealth());

    let isheadless = false;
    let browser2 = await puppeteer.launch({
        headless: isheadless,
        //ignoreHTTPSErrors: true,
        //executablePath: 'C:\\Program Files (x86)\\Google\\Chrome\\Application\\chrome.exe',
        args: []
    });
    let page2 = (await browser2.pages())[0];`

Is it possible to add/modify more variables manually to make the browser even less detectable?
thank you!

trnj:
The most valuable part that is the bottom of this screenshot.  

There are could be issues with screen res, gl vendors, incorrect timezone.

It is not related to the plugin itself, imho.

trnj:
Btw pixel scan uses a bit old lib for detection, which is not relevant to up to date providers.

prescience-data:
@trnj I haven't looked at it myself, but I'm assuming they check the pixel overhang on this right?

### Real Browser
![image](https://user-images.githubusercontent.com/65471523/93292898-04bc8500-f82a-11ea-8341-90d02ccfb443.png)


### Puppeteer Extra Stealth 2.6.1
(Available screen size +40 pixels)
![image](https://user-images.githubusercontent.com/65471523/93292879-f5d5d280-f829-11ea-94b3-f21f1f83947f.png)

trnj:
There is a bug in 2.6.1... so wait for the next release or downgrade to 2.6.0.

Experiment with vendor gl, I think your settings are not correct. Compare with a real browser. Also, check plugins, should be chrome not chromium. 

Match your settings from the real browser with headless.

Niek:
@prescience-data yes that would be a good one to add. Although I'm not sure it's causing it in this case.

trnj:
In this case, it is detected by vendor webgl settings, if I remember correctly.  
I bypassed this test.

Niek:
> There is a bug in 2.6.1... so wait for the next release or downgrade to 2.6.0.

Can you elaborate what bug you are referring to @trnj? Is it the plugins detection (must be since that's the only thing that changed between 2.6.0 and 2.6.1)?

[...]

Issue #202, opened 2020-05-17
Page redirect/ different render when using puppeteer-extra vs without


Bobchev:
Hello, I am really grateful for the plugin, but have couple of issues using it.
I am testing it with couple websites: https://www.whatismyip.com/ , https://whatismyipaddress.com/

The code I am using

```js
// puppeteer-extra is a drop-in replacement for puppeteer,
// it augments the installed puppeteer with plugin functionality
const puppeteer = require('puppeteer-extra')

// add stealth plugin and use defaults (all evasion techniques)
const StealthPlugin = require('puppeteer-extra-plugin-stealth')
puppeteer.use(StealthPlugin())

// puppeteer usage as normal
puppeteer.launch({ headless: false }).then(async browser => {
  console.log('Running tests..')
  const page = await browser.newPage()
  await page.setViewport({ width: 1366, height: 768});
  await page.goto('https://whatismyipaddress.com/')
  await page.waitFor(5000)
  await page.screenshot({ path: 'testresult.png', fullPage: true })
  await browser.close()
  console.log(`All done, check the screenshot. ✨`)
})
```

Without stealth :
![testresult](https://user-images.githubusercontent.com/12601873/82154922-d0d05380-9879-11ea-855c-9ee3af03a920.png)

With Stealth :
![testresultWithExtra](https://user-images.githubusercontent.com/12601873/82154937-e80f4100-9879-11ea-9867-7ef27b8200be.png)

It seems that I get redirect to page with AdSense ad. When I comment 
//puppeteer.use(StealthPlugin())
It is working normally.
Quite strange, can I get a hand on this issue, thank you

cam-perry:
I have the same issue on a different site which is resolved either by not using the stealth plugin or by using the adblocker plugin in conjuction. However, with the adblocker plugin enabled I run into #90/#91 and the suggestions on those issues (using version 2.4.5 or disabled accept-language evasion) do not resolve the stream of error messages.

kensnyder:
@Bobchev I experienced the same problem. I removed the evasions one at a time and found that `iframe.contentWindow` causes the problem. Namely it finds the first iframe in the document and navigates to it. Since ads are often served through iframes, you'll often see ads. If you launch with `headless: false` you can watch it happen.

So you'll need to disable the evasion. I think there is a way you can disable a single evasion, but I did this:
```javascript
puppeteer.use(stealthPlugin({
	enabledEvasions: [
		'chrome.runtime',
		'console.debug',
[...]

Issue #137, opened 2020-01-10
iframe.contentWindow evasion causes troubles for some ad systems


methyl:
Hello!

I've encountered a problem with `iframe.contentWindow` evasion. It seems to be breaking some of the websites that are displaying ads and I guess it's related to `srcdoc` attribute. With plugin enabled, the ad script instead of replacing iframe HTML content replaces whole page HTML.

You can see this being a case for eg. http://veganliftz.com/best-vegan-pre-workout.

Once I `pluginStealth.enabledEvasions.delete('iframe.contentWindow')` or enable adblocker, it starts to work again.

Niek:
Probably related to https://github.com/puppeteer/puppeteer/issues/1106

berstend:
As this is becoming the canonical ticket for this: The `iframe.contentWindow` evasion attempts to fix a nasty puppeteer bug but that comes at the expensive of some sites (specifically ad-infested ones) breaking.

Most often you don't need this evasion and can just disable it if you experience issues:

```js
pluginStealth.enabledEvasions.delete('iframe.contentWindow')
```

Phyks:
As per https://github.com/berstend/puppeteer-extra/issues/137#issuecomment-685680787, discarding this evasion or using an adblocker for websites full of ads might not be an option. Sometimes you **do** need the ads (e.g. for correct placements / layouts).

Would you have some opinion on the patch proposed above? https://github.com/berstend/puppeteer-extra/issues/202#issuecomment-645300106

Thanks!

trnj:
I've probably solved this issue and rewrote the plugin a bit. Need to test a bit. 

How does the check work? This way?
` /Chrome/.test(fingerprint.userAgent) && fingerprint.iframeChrome === 'undefined'
`
Right?

Niek:
> I've probably solved this issue and rewrote the plugin a bit. Need to test a bit.
> 
> How does the check work? This way?
> `/Chrome/.test(fingerprint.userAgent) && fingerprint.iframeChrome === 'undefined'`
> Right?

You can see the existing tests in https://github.com/berstend/puppeteer-extra/blob/master/packages/puppeteer-extra-plugin-stealth/evasions/iframe.contentWindow/index.test.js

Run the tests locally: ``npx ava -v``

trnj:
> Would you have some opinion on the patch proposed above? [#202 (comment)](https://github.com/berstend/puppeteer-extra/issues/202#issuecomment-645300106)

@Niek 
[...]

Issue #318, opened 2020-09-14
iframe.contentWindow evasion can leak Object.apply in stack traces


prescience-data:
Similar to #209 there is instances in the wild searching for: `Object.apply (<anonymous>` in thrown error stack traces.

#### Real browser:

```js
Error: Failed to execute 'createElement' on 'Document': The tag name provided ([redacted]) is not a valid name.
    at test.js?660921:2
```

#### Puppeteer with Extra Stealth 2.6.1

```js
Error: Failed to execute 'createElement' on 'Document': The tag name provided ([redacted]) is not a valid name.
    at Object.apply (<anonymous>:80:31)
    at test.js?103622:2
```

Second case fails the vendor detection test.

momala454:
careful to not remove all Object.apply, which could be detected by purposely creating a stack containing Object.apply

Niek:
Sample test code:
```js
try {
    document["createElement"](0);
} catch (e) {
    console.log(e.stack)
}
```

prescience-data:
> Sample test code:
> 
> ```js
> try {
>     document["createElement"](0);
> } catch (e) {
>     console.log(e.stack)
> }
> ```

Sample code including the observed detection:

``js
var detection;
var stack;
try {
    document.createElement(0);
} catch (err) {
    try {
        stack = err.stack.split('\n');
        detection = (stack.length >= 2) ? !!stack[1].match(/Ob[cej]{3}t\.a[lp]{3}y[\(< ]{3}an[oynm]{5}us>/) : 'Normal stack trace.';
    } catch (errOfErr) {
        detection = 'Stack trace hijacked.';
    }
}
```

berstend:
This is leveraging the ancient/shitty `iframe.contentWindow` evasion, which hasn't been updated yet to use the new stealth utilities: https://github.com/berstend/puppeteer-extra/blob/12fe3afd927327f7624e6ac0b255db02c586ac4d/packages/puppeteer-extra-plugin-stealth/evasions/iframe.contentWindow/index.js#L95

berstend:
Workaround for the time being: Disabling the `iframe.contentWindow` evasion

berstend:
Fixed in `puppeteer-extra-plugin-stealth@2.6.5`

PR #327, opened 2020-09-16
Fix iframe.contentWindow.document issues

Niek:
This should fix the issues seen in e.g. Doubleclick tags.

Thanks to: https://github.com/berstend/puppeteer-extra/issues/202#issuecomment-645300106 https://github.com/berstend/puppeteer-extra/issues/137#issuecomment-693435327 https://github.com/berstend/puppeteer-extra/issues/324#issuecomment-693452497 @phyks @trnj

Note that the test fails because `iframe.srcdoc` setter is not working - has it ever worked?

### DIFF
--- a/packages/puppeteer-extra-plugin-stealth/evasions/iframe.contentWindow/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/iframe.contentWindow/index.js
@@ -37,6 +37,10 @@ class Plugin extends PuppeteerExtraPlugin {
               if (key === 'self') {
                 return this
               }
+              // iframe.contentWindow.document === iframe.contentDocument // must be true
+              if (key === 'document') {
+                return iframe.contentDocument || iframe.contentWindow.document
+              }
               // iframe.contentWindow.frameElement === iframe // must be true
               if (key === 'frameElement') {
                 return iframe
--- a/packages/puppeteer-extra-plugin-stealth/evasions/iframe.contentWindow/index.test.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/iframe.contentWindow/index.test.js
@@ -7,9 +7,8 @@ const {
   vanillaPuppeteer,
   addExtra
 } = require('../../test/util')
-// const Plugin = require('.')
 // NOTE: We're using the full plugin for testing here as `iframe.contentWindow` uses data set by `chrome.runtime`
-const Plugin = require('puppeteer-extra-plugin-stealth')
+const Plugin = require('../..')
 
 // Fix CI issues with old versions
 const isOldPuppeteerVersion = () => {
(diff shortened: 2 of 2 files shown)


trnj:
This is interesting issue. 

It arises because Adsense does not create iframe immediately.

```
 if (!iframe.contentWindow) {
                    const proxy = new Proxy(window, contentWindowProxy)
                    Object.defineProperty(iframe, 'contentWindow', {

```

Sometimes, iframe.contentWindow would be empty in Adsense. In theory, we could await it. But I've never seen this issue using the proposed fix and probably would be enough. 

Thank you for your excellent work!

trnj:
> Note that the test fails because `iframe.srcdoc` setter is not working - has it ever worked?

Hello.

I think srcdoc should be "/**/", right?

iframe.srcdoc = "/**/"

Niek:
> > Note that the test fails because `iframe.srcdoc` setter is not working - has it ever worked?
> 
> Hello.
> 
> I think srcdoc should be "/**/", right?
> 
> iframe.srcdoc = "/**/"

What do you mean, in the test case? Can you link the line number?

trnj:
@Niek 

2.6.1 fails on this check 

```
var ifr = document.createElement('iframe');
ifr.srcdoc = "/**/";
ifr.setAttribute("style", "display: none;");

document.head.appendChild(ifr);

var descr = Object.getOwnPropertyDescriptors(HTMLIFrameElement.prototype);

if (navigator.userAgent.indexOf("Chrome") > -1 && 'function get contentWindow() { [native code] }' === descr.contentWindow.get.toString())
{
console.log('Bot has been detected!!');
}

```

When we call `descr.contentWindow.get.toString()`

Real browser
```
function()
    {
      let contentWindow = getContentWindow(this);
      injectIntoContentWindow(contentWindow);
      return contentWindow;
    }
```

Bot:
`function get contentWindow() { [native code] }
`

trnj:
@Niek 

What is the purpose of this evasion? If detect just run this code, it could detect immediately. 

```
const descriptors = Object.getOwnPropertyDescriptors(HTMLIFrameElement.prototype

if (descriptors.contentWindow.get.toString() === 'function get contentWindow() { [native code] }')
console.log('You are bot');
else
console.log(descriptors );
```

trnj:
Ok, my mistake. An extension in my browser changes HTMLIFrameElement.prototype.  Sorry, ignore. But at least double-checked that above evasion works correctly :)

berstend:
`srcdoc` as a storied history of being buggy with pptr, I'd need to do some testing to see the current status of things.

It would also be interesting to compare with playwright as AFAIK this should be fixed through their more robust context implementation (which also affected `window.open`).

Relevant issues:
https://github.com/puppeteer/puppeteer/issues/4960
https://github.com/puppeteer/puppeteer/issues/1106#issuecomment-359313898

berstend:
I'm gonna look into this PR (and `srcdoc`) soon, will probably also update the evasion to use the new stealth utilities (#318).

berstend:
Made obsolete by #318

Issue #262, opened 2020-07-19
Using with puppeteer-core?


lanly-dev:
Is it possible to use puppeteer-extra-plugin-stealth features with puppeteer-core? I assumed it can't from the look of it but still wondering if there is any chance. Thanks

brunogaspar:
There's a few issues (closed) that mentions how to use it with puppeteer-core, have you checked those out?

https://github.com/berstend/puppeteer-extra/search?q=puppeteer-core&type=Issues

lanly-dev:
I looked over the related issue, I tried but it is just too complicated to apply in my [project](https://github.com/lanly-dev/VSCode-LMPTM) (due to this [issue](https://github.com/lanly-dev/VSCode-LMPTM/issues/6)) since a lot of Typescript and Webpack errors show up after that. I'm probably going to wait and try future releases. Thanks!

Issue #66, opened 2019-10-12
Add support for puppeteer-core


jtmilne:
Can we add a constructor option for using `puppeteer-core` instead of `puppeteer` to avoid the chromium download when connecting to a remote chromium instance?

berstend:
Fix:

> according to ["puppeteer vs puppeteer-core"](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteer-vs-puppeteer-core) [](/GoogleChrome/puppeteer/blob/HEAD@{2019-09-29T16:37:13Z}/docs/api.md#puppeteer-vs-puppeteer-core) they have two main differences...
> 
> > To sum up, the only differences between puppeteer-core and puppeteer are:
> > 
> > * puppeteer-core doesn't automatically download Chromium when installed.
> > * puppeteer-core ignores all PUPPETEER_* env variables.
> 
> You can avoid downloading chrome using the [`PUPPETEER_SKIP_CHROMIUM_DOWNLOAD`](https://github.com/GoogleChrome/puppeteer/blob/v0.10.2/docs/api.md#environment-variables) [](/GoogleChrome/puppeteer/blob/HEAD@{2019-09-29T16:37:13Z}/docs/api.md#environment-variables) environment variable during installation. This makes `puppeteer` almost identical to `puppeteer-core` with the added benefit of convenient env vars.
> 
> e.g. `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1 npm install puppeteer`

:)

bronson:
This doesn't work when someone on my team does a `git pull; yarn install`.

For me, this issue still exists.

Issue #30, opened 2019-04-03
Is there smaller version like puppeteer core only?


shirshak55:
Currently I only use pupeeteer core as puppetter is too heavy and I already have chrome installed so I use chrome-paths and puppeteer-core which works really well. Is there any way to switch from pupeeteer to puppeteer core?

jchook:
I think you can still specify `executablePath` when calling [`puppeteer.launch()`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions).

shirshak55:
@jchook but u are missing point that is we need to add puppeteer otherwise this library wont work. Puppetter is peer dependency. I want to use puppeteer core because its extremely lightweight :)

jchook:
@shirshak55 according to ["puppeteer vs puppeteer-core"](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteer-vs-puppeteer-core) they have two main differences...

> 
> To sum up, the only differences between puppeteer-core and puppeteer are:
> 
> * puppeteer-core doesn't automatically download Chromium when installed.
> * puppeteer-core ignores all PUPPETEER_* env variables.
> 

You can avoid downloading chrome using the [`PUPPETEER_SKIP_CHROMIUM_DOWNLOAD`](https://github.com/GoogleChrome/puppeteer/blob/v0.10.2/docs/api.md#environment-variables) environment variable during installation. This makes `puppeteer` almost identical to `puppeteer-core` with the added benefit of convenient env vars.

e.g.  `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1 npm install puppeteer`

shirshak55:
@jchook wow thanks :) never realized there was environment that could control whether we can download chromium or not.

berstend:
I'll look into that a bit (to see if there are more differences between core and non-core). Definitely something I'll add to the documentation. :) Keeping this issue open for now.

shirshak55:
@berstend hmm I think as jchook as said i can use env variable to make puppeteer act like puppeteer core :)

PR #303, opened 2020-08-31
[WIP] feat: Rewrite to automation-extra, Support both Playwright and Puppeteer

berstend:
### Update: [Beta versions are available for testing now](https://github.com/berstend/puppeteer-extra/issues/454)

---

### Gist

- massive rewrite of the whole project 😅 
- adds a new `automation-extra` package, which is the underlying shared foundation for both `playwright-extra` and the new `puppeteer-extra` package
- adds a new `automation-extra-plugin` (cleaner API, adds Playwright events, better type safety)
- many more things, e.g. Plugins don't listen to events themselves anymore but get called from the mother ship (we had a set of issues with e.g. too many listeners that should be fixed by that)
- much improved type safety all around

Motivation for adding playwright support: https://github.com/puppeteer/puppeteer/issues/3667

### Status

- `automation-extra` - Finished: We have plugin support for both Playwright & Puppeteer
- `playwright-extra` - Finished: It's just a small entry point to `automation-extra` to make things more clear & type safe
- `automation-extra-plugin` - Finished for now: ~~I want to add shims to make plugins supporting both pptr & playwright less verbose (e.g. pptr has `page.evaluateOnNewDocument`, playwright has `context.addInitScript`)~~

~~I haven't switched `puppeteer-extra` over to use the new `automation-extra` code base yet, but specific tests against `stealth`, `adblocker`, `recaptcha` are all green 🚀~~

### Next steps

I realized pretty late in the process of adding Playwright support that I need to make the code backwards compatible with existing plugins, so new `automation-extra-plugin` based plugins work with both `playwright-extra` and `puppeteer-extra` (the latter still supporting `puppeteer-extra-plugin`'s). 😅

- [x] ~~Replace `puppeteer-extra` with a small package using the new code in `automation-extra`~~
- [x] ~~Make sure all tests in the whole monorepo are green after that~~
- [x] ~~Port the first plugin over to support both Playwright & Puppeteer~~
- [x] ~~Add/update/improve/fix documentation where needed~~
- [ ] Final public beta version testing before release

Once that's done we should be able to push/publish this stuff without breaking anything. Afterwards I'll start porting other existing plugins to support both Playwright and Puppeteer (their APIs differ, so existing puppeteer plugins need to be modified to support both - this PR is about the underlying plugin framework itself).

In the process of porting the plugins I will extend `automation-extra-plugin` wherever it makes sense (add new unified events (e.g. `addScript`) or add some sort of shim).

### Known issues

[...]

PR #336, opened 2020-09-24
feat: use puppeteer-core instead of puppeteer

ratson:
This should address https://github.com/berstend/puppeteer-extra/issues/262 https://github.com/berstend/puppeteer-extra/issues/27 https://github.com/berstend/puppeteer-extra/issues/30 https://github.com/berstend/puppeteer-extra/issues/66 

It seems this package is the `puppeteer` types rather than its functionality to download browser.

### DIFF
--- a/packages/puppeteer-extra/package.json
+++ b/packages/puppeteer-extra/package.json
@@ -40,6 +40,7 @@
   },
   "devDependencies": {
     "@types/puppeteer": "^2.0.0",
+    "@types/puppeteer-core": "^2.0.0",
     "ava": "^2.4.0",
     "documentation-markdown-themes": "^12.1.5",
     "npm-run-all": "^4.1.5",
@@ -60,11 +61,11 @@
     "typescript": "^3.7.4"
   },
   "peerDependencies": {
-    "puppeteer": "*"
+    "puppeteer-core": "*"
   },
   "dependencies": {
     "@types/debug": "^4.1.0",
-    "@types/puppeteer": "*",
+    "@types/puppeteer-core": "*",
     "debug": "^4.1.1",
     "deepmerge": "^4.2.2"
   },
--- a/packages/puppeteer-extra/src/puppeteer.ts
+++ b/packages/puppeteer-extra/src/puppeteer.ts
@@ -1,10 +1,10 @@
 // A wildcard import would result in a `require("puppeteer")` statement
 // at the top of the transpiled js file, not what we want. :-/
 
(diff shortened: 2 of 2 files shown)


berstend:
Hey @ratson - unfortunately `@types/puppeteer-core` is severely outdated :-/

This issue should be fixed in the upcoming #303 rewrite regardless, as I've ditched the peer dependencies:
https://github.com/berstend/puppeteer-extra/pull/303/files#diff-9796a6dfb75df90c2755fd54e3641cafR54-R61

Closing this PR for now, but thanks for the initiative!

AdrKacz:
Hello, what are the options to use `puppeteer-extra` only with `puppeteer-core`? (Without downloading any browser)

bjornlll:
@AdrKacz if you have `puppeteer-core` installed alongside this package it should automagically work. However, on my side it doesn't (esbuild + puppeteer-core + this package). Will let you know if I find out what the issue is.

bjornlll:
Solved by, in esbuild, marking the models as external and manually npm installing them into the bundle.

AdrKacz:
@bjornlll can you please share the config files that you found to resolve the issue?

bjornlll:
Sure, but it likely won't be relevant. 

**sst.config.ts**
```typescript
const ingestDocFnParams: Omit<sst.aws.FunctionArgs, 'handler'> = {
    // ...
    nodejs: {
      install: {
        // This marks the the packages as 'external' to esbuild and manually npm installs them
        "jsdom": "npm:jsdom-alt@25.0.2",
        "@sparticuz/chromium": "131.0",
        "sharp": "*",
        "puppeteer-core": "^23.10.4",
        "puppeteer-extra": "^3.3.6",
        "puppeteer-extra-plugin-stealth": "^2.11.2",
      }
    },
  };
```

This solution is specific to `esbuild` + SST. A similar solution would probably work if you're using another bundler (webpack, vite, ...). Key is to exclude the packages from the bundle and npm install them instead.

AdrKacz:
I'm also working with esbuild + SST so it might help more than you think ☺️
Thank you very much, I'll give it a try !

bjornlll:
Haha! Great to hear. Then you should also know that I had to patch SST to make this work: https://github.com/bjornlll/sst. Submitted a PR to them. Not sure if/when they'll merge: https://github.com/sst/sst/pull/5270

As an alternative you can also use the `sst.aws.Function`'s `bundle` property to provide your own bundling logic.

PR #338, opened 2020-09-26
Add ElementHandle re-export from puppeteer

allynsweet:
TypeScript requires puppeteer types to be from the same package


Niek (approved):

berstend:
For some reason I can't rerun the failing tests here? 🤔

berstend:
This will be fixed in #303 (soon to be released) - I found a way to export all types without a problematic `require()` statement being added.

PR #343, opened 2020-10-02
fix(plugin-stealth): Minor mistake in option names in the UA evasion

starrify:


berstend:
Oh, good catch 😄 Let me check if I can make the adblocker tests more robust (the reason CI fails quite often)

berstend (approved):

berstend:
thanks @starrify 👍

berstend:
Published in `puppeteer-extra-plugin-stealth@2.6.3`

PR #344, opened 2020-10-02
doc update

momala454:
Small update about the filename


berstend:
I think it's clear enough for anyone looking at the command line arguments :-)

As discussed this will also not matter with the upcoming changes to the stealth plugin.

PR #358, opened 2020-10-28
feat(plugin-stealth): Add navigator.hardwareConcurrency evasion

jeroenvisser101:
This PR adds a new evasion for `navigator.hardwareConcurrency`, which is useful when running in environments where a high core count could be used to detect automation.

https://antoinevastel.com/reports/stats/osName_hardwareConcurrency_report.html

It's set to the most generic/popular value, but can be customized if required.


berstend (approved):
LGTM 👍

berstend:
Released as `puppeteer-extra-plugin-stealth@2.6.4`, thanks @jeroenvisser101 👍

PR #359, opened 2020-10-29
feat(plugin-recaptcha): Hide captcha window after solving

kfirba:
All credits of this PR goes to @nsourov.

Currently, the plugin does not close the recaptcha window after solving it. I guess that in most cases the behavior is fine.

In my situation, the recaptcha window is opening on top of a button that I need to click, and what happens is that Puppeteer clicks the mouse at the position of the button, but since the recaptcha window is open, it clicks the recaptcha window and not the desired button.

This PR simply closes the recaptcha after solving it correctly.

### DIFF
--- a/packages/puppeteer-extra-plugin-recaptcha/src/content.ts
+++ b/packages/puppeteer-extra-plugin-recaptcha/src/content.ts
@@ -115,6 +115,26 @@ export class RecaptchaContentScript {
     )
   }
 
+  private _hideChallengeWindowIfPresent(id?: string) {
+    let frame: HTMLElement | null = document.querySelector<HTMLIFrameElement>(
+      `iframe[src^='https://www.google.com/recaptcha/api2/bframe'][name^="c-${id ||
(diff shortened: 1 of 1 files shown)


kfirba:
The failing test is from the adblocker extension which is irrelevant for this PR. Any idea why?

jeroenvisser101:
That's a timeout, seems like a flaky test, unrelated to this, not sure why Google would be timing out there, maybe for some reason it never reaches networkidle0?

kfirba:
@berstend Any plans on reviewing this PR?

berstend:
The `adblocker` plugin tests are a bit flaky. I issued a re-run.

berstend (approved):
LGTM 👍 Seems pretty defensively coded as well, I appreciate that.

berstend:
Published as `puppeteer-extra-plugin-recaptcha@3.1.19`

Thanks @kfirba @nsourov 👍

PR #360, opened 2020-10-29
feat(plugin-stealth): Add webdriver fix for Chrome >= 88.0.4291.0

Niek:
To test, run:
```
PUPPETEER_EXECUTABLE_PATH=/path/to/chrome-unstable yarn run test
```

### DIFF
--- a/packages/puppeteer-extra-plugin-stealth/evasions/navigator.webdriver/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/navigator.webdriver/index.js
@@ -15,10 +15,16 @@ class Plugin extends PuppeteerExtraPlugin {
     return 'stealth/evasions/navigator.webdriver'
   }
 
+  // Pre Chrome 88.0.4291.0
   async onPageCreated(page) {
     await page.evaluateOnNewDocument(() => {
       delete Object.getPrototypeOf(navigator).webdriver
     })
+  }
+
+  // Post Chrome 88.0.4291.0
+  async beforeLaunch(options) {
+    options.args.push(`--disable-blink-features=AutomationControlled`)
   }
 }
 
--- a/packages/puppeteer-extra-plugin-stealth/evasions/navigator.webdriver/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/navigator.webdriver/index.js
@@ -23,7 +23,16 @@ class Plugin extends PuppeteerExtraPlugin {
   }
 
   // Post Chrome 88.0.4291.0
+  // Note: this will add an infobar to Chrome with a warning that an unsupported flag is set
+  // To remove this bar on Linux, run: mkdir -p /etc/opt/chrome/policies/managed && echo '{ "CommandLineFlagSecurityWarningsEnabled": false }' > /etc/opt/chrome/policies/managed/managed_policies.json
   async beforeLaunch(options) {
+    // If disable-blink-features is already passed, append the AutomationControlled switch
+    options.args.forEach(e => {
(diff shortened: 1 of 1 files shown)


berstend (approved):
@Niek sweet! also nice touch of checking if the flag is already there and mentioning the docker fix for the infobar 💪

berstend:
Published in `puppeteer-extra-plugin-stealth@2.6.3`

packages/puppeteer-extra-plugin-stealth/evasions/navigator.webdriver/index.js
```diff
@@ -15,11 +15,26 @@ class Plugin extends PuppeteerExtraPlugin {
...
+  // Pre Chrome 88.0.4291.0
   async onPageCreated(page) {
     await page.evaluateOnNewDocument(() => {
       delete Object.getPrototypeOf(navigator).webdriver
     })
   }
+
+  // Post Chrome 88.0.4291.0
+  // Note: this will add an infobar to Chrome with a warning that an unsupported flag is set
+  // To remove this bar on Linux, run: mkdir -p /etc/opt/chrome/policies/managed && echo '{ "CommandLineFlagSecurityWarningsEnabled": false }' > /etc/opt/chrome/policies/managed/managed_policies.json
```
jeroenvisser101, line 27: Nice, but should we move this to the evasion's docs? Adding this to the jsdoc might help discovery. Also, according to these docs, the path for chromium is different:

https://chromium.googlesource.com/chromium/src.git/+/master/docs/enterprise/policies.md#policy-sources

```bash
#!/bin/bash
for PRODUCT_BASE in /etc/opt/chrome /etc/chromium; do
  POLICY_PATH=${PRODUCT_BASE}/policies/managed
  mkdir -p ${POLICY_PATH}
  echo '{"CommandLineFlagSecurityWarningsEnabled": false}' > ${POLICY_PATH}/managed_policies.json
done
```

wdyt?

Niek: @jeroenvisser101 I thought about documenting it, it's a good point but I don't think it matters too much (it's more of a visual thing - unless it's possible to detect by checking the increased height of the top bar). Also the fix for Windows and macOS (if you want to document it properly) is a lot less trivial. But maybe you can make a quick PR?

The path being different for Chromium is a good one, we should probably link your shell snippet.

Niek: Bynthe way, I wish there was a way to provide/overwrite the policies with a command line flag, but it looks like that is not possible.

berstend: Would be good to have macOS/Windows examples as well

jeroenvisser101: I'm not actually sure how to set this up via macOS/Win, but am open to submit a PR for documentation + shell snippet. I will see if I can find time to check if the increased height is detectable. I think it is, but I want to confirm. I'm guessing this is only an issue for headful anyway. I think most run headless, but only because headful takes extra effort/resources to set up and run.

Niek: For macOS it requires some fiddling with Profile Manager (some kind of enterprise-y product): https://sites.google.com/a/chromium.org/dev/administrators/mac-quick-start

For Windows it's even more ridiculous: https://sites.google.com/a/chromium.org/dev/administrators/windows-quick-start

PR #372, opened 2020-11-11
Webgl evasion update

kbourro:
Update webgl to use sessions, so each browser will have sticky fingerprint. Plus many more parameters added.

### DIFF
--- a/packages/puppeteer-extra-plugin-stealth/evasions/webgl.vendor/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/webgl.vendor/index.js
@@ -4,9 +4,124 @@ const { PuppeteerExtraPlugin } = require('puppeteer-extra-plugin')
 
 const withUtils = require('../_utils/withUtils')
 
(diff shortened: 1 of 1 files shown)


Niek:
In order for this to get merged, the tests need to be changed. It would also be nice if `utils.replaceProperty` can be used.

berstend:
Unfortunately the raw code is just 20% of open-source work, this needs:
- additional tests showing that properties are being spoofed
- the possibility to use plugin options to provide specific data
- linting, this doesn't seem to be formatted correctly
- static assets should be offloaded to .json files
- this needs to use stealth utilities and be tested against leaks
- the comments of old code need to be removed
- the structure of the code is pretty bad, ideally this would be cleaned up

This is currently PoC code, which is fine for internal projects but needs to be cleaned up quite a bit before we can merge this in. :-)

jjsaunier:
@kbourro if you want a bigger list of graphic driver : http://pastebin.fr/72586 my 2 cents

Niek:
Nice changes, as @berstend said it would be good to put the list of graphic drivers in a separate .json file for easier maintenance. Also, the tests are still failing, see https://github.com/berstend/puppeteer-extra/pull/372/checks.

You can run `ava -v` in the dir to run the tests manually. They are defined in `index.test.js`.

berstend:
Closing this for now. The idea is not bad but let's do it once #303 has landed.

PR #377, opened 2020-11-19
feat(plugin-recaptcha): Improve support for invisible reCAPTCHAs

### LINKED ISSUE #162: Callback not issuing correctly
Every time i use the pluguin, the callback is being transmited correctly but the webpage just ¿reloads?
Ataching video and code so you can try to help me out

```js
await page.goto(link, {waitUntil: 'load', timeout: 0})
await page.screenshot({path: 'screenshot.png'});

await page.solveRecaptchas().catch(function(error) {
    console.error(error);
});

await new Promise(r => setTimeout(r, 5000)).catch(function(error) {
    console.error(error);
});

await evaluateFirstPage(link).catch(function(error) {
    console.error(error);
});
```

https://vimeo.com/395266602

berstend:
This PR:
- adds improved support for detecting (and solving) invisible reCAPTCHAs, especially if the challenge is programmatically invoked
  - In addition to looking for visible reCAPTCHA iframes we're now looking for invisible ones as well, but filter them by checking if there's an active challenge window open for them (otherwise we would get false positives and duplicates)
- adds an additional test making sure the plugin doesn't throw in case no captchas are found

Reference:
https://developers.google.com/recaptcha/docs/invisible#programmatic_execute

Potentially fixes #162

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ yarn-error.log*
 TODO.md
 packages/testing/
 packages/puppeteer-extra2/
+packages/puppeteer-extra-old/
 packages/test-*
 
 testing/
--- a/packages/puppeteer-extra-plugin-recaptcha/src/content.ts
+++ b/packages/puppeteer-extra-plugin-recaptcha/src/content.ts
@@ -154,7 +154,7 @@ export class RecaptchaContentScript {
   }
 
   private getVisibleIframesIds() {
-    // Find all visible recaptcha boxes through their iframes
+    // Find all regular visible recaptcha boxes through their iframes
     return this._findVisibleIframeNodes()
       .filter(($f) => this._isVisible($f))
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -26,6 +26,7 @@ yarn-error.log*
 TODO.md
 packages/testing/
 packages/puppeteer-extra2/
+packages/puppeteer-extra-old/
```
Niek, line 29: Was this added by accident?

Niek (approved):

.gitignore
berstend: Nah, I'm working on the playwright rewrite and that folder is excluded in the new branch already :D

berstend: I could have selectively committed things instead but was lazy :)

berstend:
Published as `puppeteer-extra-plugin-recaptcha@3.2.1`

PR #379, opened 2020-11-21
Catch uncaught exceptions in _onTargetCreated

cjroebuck:
Getting lots of uncaught exceptions running in production when puppeteer-extra plugins try to access a page/target that is created and then destroyed instantly (for example if browser is redirected), this change will catch those exceptions. 
e.g the puppeteer-extra-recaptcha plugin calls `page.setBypassCSP` inside its onPageCreated event handler, but sometimes the page is no longer available, leading to lots of `Protocol error (Page.setBypassCSP): Target closed.` uncaught exceptions.

### DIFF
--- a/packages/puppeteer-extra-plugin/src/index.ts
+++ b/packages/puppeteer-extra-plugin/src/index.ts
@@ -531,13 +531,17 @@ export abstract class PuppeteerExtraPlugin {
    * @private
    */
   async _onTargetCreated(target: Puppeteer.Target) {
-    if (this.onTargetCreated) await this.onTargetCreated(target)
-    // Pre filter pages for plugin developers convenience
-    if (target.type() === 'page') {
-      const page = await target.page()
-      if (this.onPageCreated) {
-        await this.onPageCreated(page)
+    try {
+      if (this.onTargetCreated) await this.onTargetCreated(target)
+      // Pre filter pages for plugin developers convenience
+      if (target.type() === 'page') {
+        const page = await target.page()
+        if (this.onPageCreated) {
+          await this.onPageCreated(page)
+        }
       }
+    } catch (err) {
+      this.debug(`_onTargetCreated: caught error: ${err.message}`}
     }
   }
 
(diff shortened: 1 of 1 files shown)


berstend:
I've never experienced this. Can you make a test case where this behavior is shown?

cjroebuck:
> I've never experienced this. Can you make a test case where this behavior is shown?

Will try to, might be something I'm doing on top that is causing this behaviour

> Also it might more sense to do this in the `onPageCreated` handlers themselves?

Perhaps but I thought this would just catch all exceptions in one place rather than having to add it to each plugin separately.

> Is only page.setBypassCSP an issue here or other page functions as well?

It would be any access to `page` object, once the target is closed.

berstend:
Would be good to get clarification on what exactly is causing the errors to be thrown in your setup. :-)

Errors being thrown is not a bad thing, as it indicates that something went wrong. 

I'm personally not a huge fan of those godly try/catch blocks: e.g. if something went wrong in a `onPageCreated` handler the developer wouldn't know unless they enabled debug mode

If this is a general issue (and not caused by something in your setup) we could e.g. check specifically for `Target closed` errors and swallow those and re-throw everything else.

cjroebuck:
hehe agree, it's just I'm getting tons on uncaught exceptions from this plugin and have no control over catching them as their in a separate async scope to my own code. I'll try to get a test case.

berstend:
Got it :-) Haven't experienced or heard about widespread `Target closed` issues in conjunction with the recaptcha plugin - hence it'd be good to understand the core issue better. :-)

berstend:
As an ugly quick fix for the moment: You could add an [`uncaughtException` handler](https://nodejs.org/api/process.html#process_warning_using_uncaughtexception_correctly)? Still curious to learn what is causing this though.

cjroebuck:
Yep I have the handler and I have alerts whenever it gets called in production, hence me wanting to silence it! I'll take a deeper look into what's causing it and come back

cjroebuck:
🤦🏻‍♂️ Yep, it was because I had another bit of code that was closing pages that were opened from another page

berstend:
@cjroebuck Haha, no worries :-) Thanks for digging into this and reporting back 👍

PR #400, opened 2021-01-12
feat(plugin-recaptcha): Add support for reCAPTCHA Enterprise Add reCAPTCHA Enterprise support

ev-kenjiterai:
Hi,

Thanks for maintain library.
While using puppeteer-extra-plugin-recaptcha, I found another set of urls for reCAPTCHA enterprise.

I added urls to recaptcha find method.

If you can, could you merge that?
I don't know enough your community rules, tell me what should I do if wrong.

thanks.

### DIFF
--- a/packages/puppeteer-extra-plugin-recaptcha/src/content.ts
+++ b/packages/puppeteer-extra-plugin-recaptcha/src/content.ts
@@ -115,6 +115,8 @@ export class RecaptchaContentScript {
     return Array.from(
       document.querySelectorAll<HTMLIFrameElement>(
         `iframe[src^='https://www.google.com/recaptcha/api2/anchor'][name^="a-"]`
+        + ', ' +
+        `iframe[src^='https://www.google.com/recaptcha/enterprise/anchor'][name^="a-"]`
       )
     )
   }
@@ -123,12 +125,20 @@ export class RecaptchaContentScript {
       `iframe[src^='https://www.google.com/recaptcha/api2/anchor'][name^="a-${
         id || ''
       }"]`
+      + ', ' +
+      `iframe[src^='https://www.google.com/recaptcha/enterprise/anchor'][name^="a-${
+        id || ''
+      }"]`
     )
   }
 
   private _hideChallengeWindowIfPresent(id?: string) {
     let frame: HTMLElement | null = document.querySelector<HTMLIFrameElement>(
       `iframe[src^='https://www.google.com/recaptcha/api2/bframe'][name^="c-${
+        id || ''
+      }"]`
+      + ', ' +
+      `iframe[src^='https://www.google.com/recaptcha/enterprise/bframe'][name^="c-${
         id || ''
       }"]`
     )
@@ -183,6 +193,10 @@ export class RecaptchaContentScript {
         (id) =>
           document.querySelectorAll(
             `iframe[src^='https://www.google.com/recaptcha/api2/bframe'][name^="c-${
+              id || ''
+            }"]`
+            + ', ' +
+            `iframe[src^='https://www.google.com/recaptcha/enterprise/bframe'][name^="c-${
               id || ''
             }"]`
           ).length


Niek (approved):

brunogaspar (approved):

berstend:
@ev-kenjiterai thanks for the PR! Did you have a chance to test if this works?

Specifically I'm wondering if we need to pass a `version=enterprise` to the 2captcha API for enterprise captchas to be solved correctly: https://2captcha.com/2captcha-api#solving_recaptcha_enterprise

2captchas API documentation is notoriously bad, so this might not be required in practice.

ev-kenjiterai:
Hi, @berstend 

Thanks, advice!
I used that with my work, it just work.
But, I will test again carefully with `version=enterprise` to the 2captcha API.

I will update soon!

berstend:
@ev-kenjiterai sounds good! I just want to ensure we don't break regular recaptcha functionality. :-)

Is there a known demo site for enterprise recaptchas or a public site we could use for automated tests?

berstend:
A discord member confirmed this patch makes the recaptcha plugin work on Netflix.

berstend:
```
Successfully published:
 - puppeteer-extra-plugin-recaptcha@3.3.2
```

divinedeveloper:
@ev-kenjiterai @berstend some recaptcha enterprise like that on [Hulus Login page](https://www.hulu.com/welcome) are not solved even after trying out `page.solveRecaptchas()` and `frame.solveRecaptchas()`  with iframes. Do we need to set something like `version=enterprise` to RecaptchaPlugin provider like the way to pass token or anything else to auto detect enterprise recaptchas? Thanks in advance.

berstend:
@divinedeveloper in my (limited) testing with enterprise setting that flag wasn't necessary - could you post some debug output (mentioned in the readme) here if you got the time? Curious if the detection or the solving part is the issue here

divinedeveloper:
@berstend When we solve recaptcha using the `enterprise: 1` in the `decodeReCaptcha` function, the success rate is higher for login into the hulu site, Also I am attaching the image for 2captcha where we can see that the captchatype is set as  `recaptcha_enterprise` when we set the above enterprise:1 flag explicitly. With the current default behavior you can see that it sets the catchatype as `recaptcha_v2`.

Can you please help add an `enterprise flag as plugin option`, as we have validated that it sets captchatype  as `recaptcha_enterprise`

2captcha-hulu

berstend:
@divinedeveloper thanks for the data point that the enterprise flag increases your success rate (it didn't make a difference in my testing, hence I didn't add support for it :-))

Will see/test how to best detect an enterprise deployment and update the recaptcha plugin to use the enterprise flag in those situations. Should be able to do that within the next two days :-)

divinedeveloper:
@berstend Thanks for quick turnaround. Really appreciated

divinedeveloper:
Hey @berstend did you get any chance for releasing this fix? Thanks

berstend:
@divinedeveloper I'm working on it in conjunction with #482

We need proper testing for all the different recaptcha deployment types as this starts to get messy 😄 

I've started setting up demo pages with all the different recaptcha deployment types here: https://github.com/berstend/puppeteer-extra/pull/482#issuecomment-825561702

Will add a recaptcha enterprise demo as well (they have a free tier), so we have a proper testing target before I finish up adding 2captcha enterprise flag support. :-)

berstend:
@divinedeveloper this is so strange, I implemented the enterprise flag and in my own testing it makes things worse again:

image

berstend:
@divinedeveloper opt-in support for the enterprise flag has landed in `puppeteer-extra-plugin-recaptcha@3.3.9` (see #486) - curious to hear if it improves your success rates.

divinedeveloper:
Hey @berstend, Thanks for making this fix ready quickly. Unfortunately this fix is not working for us and we too are getting the `ERROR_CAPTCHA_UNSOLVABLE`.

I will help with the following details which we made to the `version 3.3.8` by directly tweaking the compiled version of the files inside node_modules for file `index.cjs.js` which increased the success ratio.

Please refer the attached compiled version of the patch, there we are simply passing the enterprise as 1 or 0 in the `postData` and it was perfectly working fine. Also simply added `enterpriseCaptcha: boolean;` to the PluginOptions in types.d.ts. Those where the only changes we made.

[patch.zip](https://github.com/berstend/puppeteer-extra/files/6376164/patch.zip)

Thanks again.

berstend:
@divinedeveloper that's really strange, but it kinda hints that the issue in my testing is not with the recaptcha flag but with something else (which is good in a sense).

There's not much difference between your patch and what v3.3.10 is doing:
```js
    if (opts.useActionValue && captcha.action) {
      extraData['action'] = captcha.action // Optional v3/enterprise action
    }
    if (opts.useEnterpriseFlag && captcha.isEnterprise) {
      extraData['enterprise'] = 1
```

Except for the new action value passing (which is opt-out) the API data should effectively be the same 🤔

berstend:
To figure this out (and see what the effective difference is) it'd be necessary to print and compare `postData` before `request.write(postData);` in both versions, I have to take care of other things but try to look into this later today.

berstend:
@divinedeveloper are you sure your version worked? 🤔 

I've pointed your patched 3.3.8 version at https://berstend.github.io/static/recaptcha/enterprise-checkbox-explicit.html and get `ERROR_CAPTCHA_UNSOLVABLE` as well.

API wise the only difference between that version and the current latest plugin version:
- no `action` parameter
- positioning of query params: it's `&enterprise=1&googlekey=foo` and not `&googlekey=foo&enterprise=1`

I don't see how that would make any difference 🤔 

The issue remains that `&enterprise=1` causes unsolvable captchas - will look more into that but unfortunately can't use your patched version as reference.

berstend:
@divinedeveloper
I did some more testing and I'm deeply confused the enterprise flag (and your patched version) work for you 😄

If it weren't for that I'd be leaning towards the 2captcha API being broken and the enterprise flag simply not working as expected.

Demo code (using `v3.3.8` + your patch, using hulu.com just in case):
```js
const puppeteer = require("puppeteer-extra")

const RecaptchaPlugin = require("puppeteer-extra-plugin-recaptcha")
puppeteer.use(
  RecaptchaPlugin({
    provider: {
      id: "2captcha",
      token: process.env.TWOCAPTCHA_TOKEN,
    },
    enterpriseCaptcha: true,
  })
)

puppeteer
  .launch({ headless: false, defaultViewport: null })
  .then(async (browser) => {

    const url = "https://www.hulu.com/welcome"
    await page.goto(url, { waitUntil: "networkidle0" })
    await page.waitForTimeout(5 * 1000)
    console.log("Finished loading")

    await page.click(`button[href*="/login"]`)
    console.log("Triggered login box")
    await page.waitForTimeout(5 * 1000)
    
    console.log("Attempting to solve captcha")
    let res = await page.solveRecaptchas()
    console.log(res)

    console.log("All done")
    await browser.close()
  })
```

Using `enterpriseCaptcha: true`:
```
method=userrecaptcha&key=2CAPTCHA_KEY&soft_id=2589&pageurl=https%3A%2F%2Fwww.hulu.com%2Fwelcome&googlekey=6LcCpMMZAAAAAElKx17CMCRCUCdoixeXgG5PRAg-&enterprise=1
```
==> `ERROR_CAPTCHA_UNSOLVABLE`

Using `enterpriseCaptcha: false`:
```
method=userrecaptcha&key=2CAPTCHA_KEY&soft_id=2589&pageurl=https%3A%2F%2Fwww.hulu.com%2Fwelcome&googlekey=6LcCpMMZAAAAAElKx17CMCRCUCdoixeXgG5PRAg-
```
==> `Captcha solved`

The only difference being `&enterprise=1` added.

I checked the 2captcha API multiple times and theoretically this should work, but it doesn't.

Can you verify the result of the above script when ran against your modified version?

To add the API parameter logging just search for `request.write(postData);` in `node_modules/puppeteer-extra-plugin-recaptcha/dist/index.cjs.js`

divinedeveloper:
@berstend  Thanks for the above details. We too find it strange and surprising that our fix has suddenly stopped working for us from today on our local too, We think something has been changed at 2captcha side and need to dig more into it. I will keep you posted what we could find from our research on 2captcha.

Till then is it possible for you to create an hulu account and test it without an patch or fixes, because it works one or 2 times but than gets blocked. Thanks

berstend:
I'd be happy to improve the recaptcha plugin whenever possible, but I don't think we or I can fix the 2captcha enterprise flag issues.

The default way of not using the enterprise flag works to solve the recaptcha, so I'm leaving it at that until there's more concrete info on how to improve things.

> create an hulu account and test it without an patch or fixes, because it works one or 2 times but than gets blocked

I'm not sure what "works" means in this context - as long as the captcha is correctly detected, it's info extracted, passed to 2captcha and returned to the site correctly that's pretty much all the recaptcha plugin can do. 🤔

PR #405, opened 2021-01-21
fix(plugin-user-data-dir): Prevent ENOTEMPTY error

urlbox-io:
Without the change, `fse.removeSync` only works with empty folder, so fails with ENOTEMPTY - https://github.com/jprichardson/node-fs-extra/issues/178


berstend (approved):
Thanks @urlbox-io 👍

berstend:
```
Successfully published:
 - puppeteer-extra-plugin-user-data-dir@2.2.9
```

PR #410, opened 2021-01-29
feat(automation-extra): adding enhanced ambient types for puppeteer

prescience-data:
- Added unexposed types such as `page._client` to puppeteer definitions
- Exported CDP command definitions (added `devtools-protocol` to devDependencies)
- Upgraded and sorted package.json with `npx sort-package-json` and `npx npm-check-updates`
- Added .idea to .gitignore
- Set puppeteer, playwright-core to peer dependencies

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ dist
 .env.production.local
 .cache
 .rpt2_cache
+.idea
 
 lerna-debug.log*
 lerna-error.log*
--- a/README.md
+++ b/README.md
@@ -48,6 +48,8 @@ yarn bootstrap
 yarn build
 ```
 
+> **Note regarding `fsevents`:** If you are working from a non-OSX environment you may need to run `npx json -I -f package.json -e 'this.resolutions={}'` prior to the initial `yarn` command to allow correct installation.
+
 #### Development flow
 
 ```bash
--- a/package.json
+++ b/package.json
@@ -1,47 +1,45 @@
 {
+  "version": "0.0.0",
   "private": true,
   "description": "Modular framework to teach Puppeteer new tricks.",
   "repository": "berstend/puppeteer-extra",
-  "author": "berstend",
   "license": "MIT",
+  "author": "berstend",
   "main": "packages/puppeteer-extra/index.js",
-  "engines": {
-    "node": ">=8"
-  },
+  "workspaces": [
+    "packages/*"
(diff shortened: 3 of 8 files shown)


prescience-data:
😭 TypeScript, why must you resist meeeee....

berstend:
> 😭 TypeScript, why must you resist meeeee....

Haha, welcome to the wonderful word of monorepo development where multiple packages use and import the same types, which makes them very sensible to version mismatches 😄

I haven't checked how `npx npm-check-updates` does things but to reliably update dependencies in all packages at the same time (to prevent typing mismatches as you've run into) `yarn lernaupdate --lazy` should be used.

Overall updating dependencies in a monorepo is a delicate endeavor, as that can introduce subtle changes in behavior or bugs so thorough testing and changelog review is needed (worst case: breaking API changes that require updating code).

Doing that at the same time as running `npx sort-package-json` (which should be run in all packages btw, maybe using `yarn lerna exec`) might not be the best thing, as it makes review much harder.

There's a reason we pinned resolutions to `"fsevents": "^2.1.2"`, I forgot what it was (and we should verify that root cause to do that is still present) but the change in this PR is most likely not what we want (the `npx json -I -f package.json -e 'this.resolutions={}'` command was a hack for CI servers and is not a real solution, as it permanently modifies the `package.json`).

Regarding exposing private puppeteer types: For `_createPageInContext` there was an immediate reason, as we're using that to patch necessary things in the plugin framework. I'm not a huge fan of shotgun-exposing a bunch more private APIs which we currently don't use (there's a reason for them to be private after all).

Regarding `peerDependencies`: There's a long story here but removing them was a conscious choice, as they don't really do what we need them for.

I also rather not use `_client`, but use a custom shim (with a dedicated API) which supports both pptr and playwright. :-)

Overall I appreciate the initiative but it'd be better to discuss the need for changes first, before working on a PR. :-) The `automation-extra` branch is still in flux and I'd like to get it to a stable point first, before doing cosmetic changes (ideally in smaller easy-to-reason about PRs).

berstend:
I've added `.idea` exclusion and the fsevents hint here: [`8ec11bc` (#303)](https://github.com/berstend/puppeteer-extra/pull/303/commits/8ec11bce076666d63ee04ee5738b2f4b8916a201)

Closing the PR for now as per my previous comment :-)

PR #413, opened 2021-01-30
feat(plugin-stealth): Add support for UA hints

Niek:
This PR adds support for UA hints to the `user-agent-override` evasion. Note that this is a breaking change, I opted to change the functionality of the evasion a bit and detect the platform (as well as the now required version, model, architecture, etc) all from the provided user agent string.

Unfortunately UA hints only work in headful (yet another reason not to use headless...), so the test needs to run in headful.

### DIFF
--- a/packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.js
@@ -115,7 +115,7 @@ class Plugin extends PuppeteerExtraPlugin {
     // Return OS version
     const _getPlatformVersion = () => {
       if (ua.includes('Mac OS X')) {
-        return ua.match(/Mac OS X ([^_]+)/)[1]
+        return ua.match(/Mac OS X ([^)]+)/)[1]
       } else if (ua.includes('Android')) {
         return ua.match(/Android ([^;]+)/)[1]
       } else if (ua.includes('Windows')) {
@@ -130,7 +130,7 @@ class Plugin extends PuppeteerExtraPlugin {
 
     // Return the Android model, empty on desktop
     const _getPlatformModel = () =>
-      _getMobile() ? ua.match(/Android.*?;.*?\/([^;]+)/)[1] : ''
+      _getMobile() ? ua.match(/Android.*?;\s([^)]+)/)[1] : ''
 
     const _getMobile = () => ua.includes('Android')
 
--- a/packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.js
@@ -7,6 +7,7 @@ const { PuppeteerExtraPlugin } = require('puppeteer-extra-plugin')
  *
  * If you don't provide any values this plugin will default to using the regular UserAgent string (while stripping the headless part).
  * Default language is set to "en-US,en", the other settings match the UserAgent string.
+ * If you are running on Linux, it will mask the settins to look like Windows. This behavior can be disabled with the `maskLinux` option.
  *
  * By default puppeteer will not set a `Accept-Language` header in headless:
  * It's (theoretically) possible to fix that using either `page.setExtraHTTPHeaders` or a `--lang` launch arg.
@@ -35,6 +36,7 @@ const { PuppeteerExtraPlugin } = require('puppeteer-extra-plugin')
  * @param {Object} [opts] - Options
  * @param {string} [opts.userAgent] - The user agent to use (default: browser.userAgent())
  * @param {string} [opts.locale] - The locale to use in `Accept-Language` header and in `navigator.languages` (default: `en-US,en`)
+ * @param {boolean} [opts.maskLinux] - Wether to hide Linux as platform in the user agent or not - true by default
  *
  */
 class Plugin extends PuppeteerExtraPlugin {
@@ -55,15 +57,24 @@ class Plugin extends PuppeteerExtraPlugin {
   get defaults() {
     return {
       userAgent: null,
-      locale: 'en-US,en'
+      locale: 'en-US,en',
+      maskLinux: true
     }
   }
 
   async onPageCreated(page) {
     // Determine the full user agent string, strip the "Headless" part
-    const ua =
+    let ua =
       this.opts.userAgent ||
       (await page.browser().userAgent()).replace('HeadlessChrome/', 'Chrome/')
+
+    if (
+      this.opts.maskLinux &&
+      ua.includes('Linux') &&
+      !ua.includes('Android') // Skip Android user agents since they also contain Linux
+    ) {
+      ua = ua.replace(/\(([^)]+)\)/, '(Windows NT 10.0; Win64; x64)') // Replace the first part in parentheses with Windows data
+    }
 
     // Full version number from Chrome
     const uaVersion = ua.includes('Chrome/')
--- a/packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.js
@@ -40,10 +40,16 @@ const { PuppeteerExtraPlugin } = require('puppeteer-extra-plugin')
 class Plugin extends PuppeteerExtraPlugin {
   constructor(opts = {}) {
     super(opts)
+
+    this._headless = false
   }
 
   get name() {
     return 'stealth/evasions/user-agent-override'
+  }
+
+  get dependencies() {
+    return new Set(['user-preferences'])
   }
 
   get defaults() {
@@ -61,8 +67,8 @@ class Plugin extends PuppeteerExtraPlugin {
 
     // Full version number from Chrome
     const uaVersion = ua.includes('Chrome/')
-      ? ua.match(/Chrome\/([^\s]+)/)[1]
-      : (await page.browser().version()).match(/\/([^\s]+)/)[1]
+      ? ua.match(/Chrome\/([\d|.]+)/)[1]
+      : (await page.browser().version()).match(/\/([\d|.]+)/)[1]
 
     // Get platform identifier (short or long version)
     const _getPlatform = (extended = false) => {
@@ -114,12 +120,12 @@ class Plugin extends PuppeteerExtraPlugin {
 
     // Return OS version
     const _getPlatformVersion = () => {
-      if (ua.includes('Mac OS X')) {
+      if (ua.includes('Mac OS X ')) {
         return ua.match(/Mac OS X ([^)]+)/)[1]
-      } else if (ua.includes('Android')) {
+      } else if (ua.includes('Android ')) {
         return ua.match(/Android ([^;]+)/)[1]
-      } else if (ua.includes('Windows')) {
-        return ua.match(/([\d|.]+);/)[1]
+      } else if (ua.includes('Windows ')) {
+        return ua.match(/Windows .*?([\d|.]+);/)[1]
       } else {
         return ''
       }
@@ -136,7 +142,6 @@ class Plugin extends PuppeteerExtraPlugin {
 
     const override = {
       userAgent: ua,
-      acceptLanguage: this.opts.locale || 'en-US,en',
       platform: _getPlatform(),
       userAgentMetadata: {
         brands: _getBrands(),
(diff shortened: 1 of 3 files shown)


packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.js
```diff
@@ -35,7 +35,6 @@ const { PuppeteerExtraPlugin } = require('puppeteer-extra-plugin')
  * @param {Object} [opts] - Options
```
berstend, line 35: The example code in the comment higher up (can't select that) still shows `platform` which we should remove and maybe add a note that it'll be derived/spoofed based on the UA string (while we're at it: we should remove `q=0.9` from the `locale` example as that isn't supported)

packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.test.js
```diff
@@ -120,22 +120,42 @@ test('stealth: navigator.languages with custom locale', async t => {
...
+
   const page = await browser.newPage()
 
-  const platform = await page.evaluate(() => navigator.platform)
-  t.true(platform === 'MyFunkyPlatform')
+  await page.goto('https://headers.cf/headers/?format=raw')
+  const firstLoad = await page.content()
+  t.true(
+    firstLoad.includes(
+      `sec-ch-ua: "Chromium";v="88", "Google Chrome";v="88", ";Not A Brand";v="99"`
```
berstend, line 149: Is there a chance `88` will be outdated once we upgrade pptr and the test will fail? 🤔

packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.test.js
```diff
@@ -120,22 +120,42 @@ test('stealth: navigator.languages with custom locale', async t => {
...
+    )
+  )
+
+  await page.reload()
+  const secondLoad = await page.content()
+  t.true(secondLoad.includes('sec-ch-ua-mobile: ?0'))
+  t.true(secondLoad.includes('sec-ch-ua-full-version: "88.0.4324.96"'))
```
berstend, line 156: Same question as in the first load test: Alternatively to hard-coding the version number we could either dynamically extract it from pptr or use a more abstract test (assuming this test will fail with a different pptr/browser version)

berstend (changes requested):
Good stuff! Added few comments/suggestions :-)

packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.js
berstend: PS: After updating jsdoc comments we should run `yarn docs-for-evasions && yarn postdocs-for-evasions` in the stealth dir

Niek: Thanks, I missed that! Fixed in 12a76f2

packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.test.js
Niek: Actually no, the version is coming from this line: https://github.com/berstend/puppeteer-extra/pull/413/files#diff-a9da9f27e487fcdb0773c1ec2bbb6a040cbd643f81fb66ecdfdf48238b6cd5cdR127

We could use a fake `99.99.99.99` version if you prefer that though.

packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.test.js
Niek: See above

packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.test.js
berstend: Ah, never mind - didn't realize the data is fully defined by us :-)

berstend (approved):

packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.test.js
Niek: Added some more tests now, used a fake version as well to be sure

berstend:
I just noticed a standard use-case we might want to optimize for:
Say I'm running stealth/pptr in docker and I just want to spoof the platform, I'd need to get the UA string myself (at a point where `page.browser().userAgent()` doesn't exist yet), string replace the platform and then pass it to the plugin, correct?

I think deriving the platform from the UA string is neat and was wondering if we can additionally provide a platform override as well (in case the rest of the UA isn't relevant to the user). Maybe there's a lightweight UA string manipulation library out there that allows us to just replace UA.platform (similar to the `URL` object)?

Niek:
> I just noticed a more common use-case we might want to optimize for:
> Say I'm running stealth/pptr in docker and I just want to spoof the platform, I'd need to get the UA string myself (at a point where `page.browser().userAgent()` doesn't exist yet), string replace the platform and then pass it to the plugin, correct?
> 
> I think deriving the platform from the UA string is neat and was wondering if we can additionally provide a platform override as well (in case the rest of the UA isn't relevant to the user). Maybe there's a lightweight UA string manipulation library out there that allows us to just replace `UA.platform` (similar to the `URL` object) which run before the rest of the code?

That's a good one, let me look into this. Another thing I thought of is skipping the `locale` parameter and converting that into the `language` preference in case headful is used - WDYT?

berstend:
> Another thing I thought of is skipping the `locale` parameter and converting that into the `language` preference in case headful is used - WDYT?

How do you mean? :-) We do set `acceptLanguage` based on the locale which affects both headless/headful?

Niek:
> > Another thing I thought of is skipping the `locale` parameter and converting that into the `language` preference in case headful is used - WDYT?
> 
> How do you mean? :-) We do set `acceptLanguage` based on the locale which affects both headless/headful?

Correct, and that leads to wrong order for the `Accept-Language` header. So in case of headful it's probably better to skip that and use the preference setting for language instead (same effect but not messing up the headers).

berstend:
Yeah, that makes sense. I'm a bit concerned as the user-data-dir plugin is quite old and not as well tested but overall it should do the trick :-) It'd be good to detect if we're using `connect`, so we can fallback to the old method to set the language header.

Niek:
> Yeah, that makes sense. I'm a bit concerned as the user-data-dir plugin is quite old and not as well tested but overall it should do the trick :-) It'd be good to detect if we're using `connect`, so we can fallback to the old method to set the language header.

I pushed some changes in 04445d5: it always sets the preferences - which are ignored on headless. On headless, it uses the `acceptLanguage` parameter. 

Good point about `connect()`, I pushed a fix for that in fe37796

berstend:
Looks good! Do we want to fix the Linux use-case before merging this in? I can't think of a scenario where it'd be desired to signal linux as the platform 🤔 

A quick fix could be:
- Detect if the UA has parenthesis is linux (search for Linux in UA string) - [examples](https://developers.whatismybrowser.com/useragents/explore/operating_system_name/linux/)
- If Linux then replace the first parenthesis with Windows 10 data
- Make this opt-out with a `transformLinux: true` option or so

WDYT?

Niek:
Hmmm, while it definitely makes sense UX-wise, it can also be confusing. I assume that many users run macOS or Windows to test and Linux to deploy - having the default of this evasion act different depending on the host is a bit strange IMHO. Maybe it makes more sense to output a big fat warning in case the UA contains `Linux`, something in the lines of "Warning: your user agent signals the Linux OS. This could be a signal to some sites to block your browsing session."

berstend:
Replacing Linux as the platform has been the default since the first version of this evasion :-)

There's still no good UX story here when upgrading to this version of the evasion and running things in a docker container (and wanting to hide that) - it'd still be necessary to provide a full UA string which is hard to do when initializing the plugin and if hardcoded will result in outdated/mismatching version numbers quickly 🤔

berstend:
Small thing, we might want to add the `user-preferences` to the stealth dependencies or things will initially fail for users

Niek:
> Replacing Linux as the platform has been the default since the first version of this evasion (in a broken way though lol) :-)

Are you sure about that? I can't find any place where "Linux" in the UA is replaced with "Windows NT XX" for example. Only the `platform` value was hardcoded to `Win32`, but the UA still signaled Linux or macOS or whatever the user has. This was initially one of my main reasons to fix the evasion - some people complained about the mismatch between the UA string and `navigator.platform` 😄 

> There's still no good UX story here when upgrading to this version of the evasion and running things in a docker container (and wanting to hide that) - it'd still be necessary to provide a full UA string which is hard to do when initializing the plugin and if hardcoded will result in outdated/mismatching version numbers quickly 🤔

Hmm yes I agree - it's kind of a breaking change. What do you suggest? There's no way to make "required" options, right?

berstend:
Apologies, I didn't formulate that clear enough: Yeah, initially we used the `anonymize-ua` plugin for this stuff, which did anonymize the UA string. Eventually we switched to using CDP methods and ended up doing it half baked (only spoofing the platform), not replacing Linux in the UA was an oversight really 😄 

Overall I think it'd be good to replace Linux (both in the UA and the platform) by default, with an opt-out (also something we didn't have before) or automatic opt-out when providing a custom UA string.
I feel that would satisfy most use-cases while staying backwards compatible with the previous approach (or at least intent though).

As for the implementation: Not sure if it's worthwhile pulling in the `anonymize-ua` plugin for this, as opposed to just copy pasting the necessary code to this evasion. Overall the stealth plugin is used the most, so limiting the number of dependencies (and things that can go wrong) might be a better approach. :-)

berstend:
I agree that masking Linux is a fair bit "opinionated" though :-)

berstend:
Let's approach this differently, the previous evasion was broken in behavior (spoofing Win32 platform on macOS as well, etc).

What I would want a proper `user-agent-override` evasion to do:
- Remove notions of `Headless` from the UA by default
- Ability to set a custom locale in the best way possible (user prefs vs CDP) 
- Ability to set a completely custom UA string
- Ability to spoof `navigator.platform` (automatically based on the UA string is a bonus)
- Ability to spoof client hints matching all the other data
- Ability to retain the original UA string/versions (sans headless) while replacing Linux with Windows (also in `navigator.platform`), with opt-out

Except for the last thing we're already there 😄 The reason I think masking Linux by default is worthwhile: Everyone running pptr in docker will have do it anyway and the current ergonomics (removing the `user-agent-override` first, etc) are pretty verbose

Niek:
OK cool, that sounds good to me. I'll add a `spoofWindows` boolean option (default true) - I'll spoof it on macOS + Linux I guess (not just Linux).

berstend:
Sounds good :-) No strong preference regarding macOS > Linux, though I think it'd be fine not spoofing macOS (as it's considered a regular user platform of sorts lol)

Niek:
> Sounds good :-) No strong preference regarding macOS > Linux, though I think it'd be fine not spoofing macOS (as it's considered a regular user platform of sorts lol)

Good point - added Linux-only spoofing now with a `maskLinux` options (default true). I excluded Android devices because if an Android UA is set that's very likely on purpose.

berstend:
Looking good 👍

berstend:
```
Successfully published:
 - puppeteer-extra-plugin-stealth@2.7.0
```

packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.js
```diff
@@ -28,43 +29,147 @@ const { PuppeteerExtraPlugin } = require('puppeteer-extra-plugin')
...
+      } else if (ua.includes('Android')) {
+        return 'Android'
+      } else if (ua.includes('Linux')) {
+        return 'Linux'
+      } else {
+        return extended ? 'Windows' : 'Win32'
+      }
+    }
+
+    // Source in C++: https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/chrome_content_browser_client.cc;l=1187-1238
```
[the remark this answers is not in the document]
Niek: Hmm yes, it even moved to a different file already... correct link is: https://source.chromium.org/chromium/chromium/src/+/master:components/embedder_support/user_agent_utils.cc;l=49-100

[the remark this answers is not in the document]
jeroenvisser101: Thanks!

berstend:
Since the `user-preferences` plugin became much more relevant with this change I added tests here: [`fdae8b0` (#303)](https://github.com/berstend/puppeteer-extra/pull/303/commits/fdae8b0af05f1601818f75e8e6596a1894de803d) (also to check everything works with the rewrite)

ushuz:
The dependency on `user-preferences` has some side effects.

`user-preferences` overrides certain user prefs on each launch, making it pretty hard to persist those prefs when needed. And since it's an inexplicit dependency, it's hard to debug either. I have to dig into each evasion's code to get to know that `user-agent-override` evasion is somehow messing my user prefs.

Since this dependency is meant to avoid initial failure https://github.com/berstend/puppeteer-extra/pull/413#issuecomment-771034822, it may be better to have an option to avoid overriding existing user prefs.

PR #418, opened 2021-02-03
fix(plugin-stealth): Fix AutomationControlled flag override

mnmkng:
This probably never affected any users, but the appending of `AutomationControlled` to existing flag would not work, because `e += ',AutomationControlled'` does not actually update the value in the `args` array, only the `e` variable inside the callback.

```js
const a = ['x','y','z'];
a.forEach(c => { c+= 'foo' } );

console.log(a) // ['x','y','z']
```

image

I also assume the `return` on line 33 was there to break out of the loop early, but again, it doesn't do anything.

This PR fixes it by inserting a new string into the array at the original index.


mnmkng:
Hah, sorry, I just copy pasted it from my own code. Did not realize there were custom changes. Will fix in a sec.

Niek:
Oops... I missed that 😅 Thanks for the PR!

Niek (approved):

berstend (approved):

berstend:
thanks @mnmkng 👍 
```
Successfully published:
 - puppeteer-extra-plugin-stealth@2.7.4
```

PR #419, opened 2021-02-03
chore(ci): Bump NodeJS to v15 and puppeteer to v7

Niek:
* Change NodeJS 14 to 15
* Change Puppeteer 5.0.0 to 7.0.0

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,10 +26,12 @@ jobs:
     strategy:
       matrix:
         node:
-          - 14
+          - 15
+          # - 14
           - 12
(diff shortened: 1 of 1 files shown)


Niek:
Seems like the recently merged UA override / hints breaks with the new Chromium of Puppeteer 7 - will investigate.

berstend:
> Update: apparently the UA hints spec got updated (header values need to be prefixed with Sec-CH-) fixed now server-side in headers.cf.

Interesting, thanks for noticing 👍

berstend (approved):

Issue #437, opened 2021-02-22
[Feature] Puppeteer Stealth does not pass geocaptcha


Edweis:
**Feature request**
`puppeteer-extra-plugin-stealth` works well and I have used it for many website, however I got caught scraping [this website](https://www.dicksmith.com.au/da/). Returning an iframe from  `https://geo.captcha-delivery.com` with:
![Gotcha](https://user-images.githubusercontent.com/18052624/108713638-162d5a00-754b-11eb-956c-6f9b12527663.png)

Any idea on how I can bypass such protection? If so I can contribute with a PR.

Thanks !

sanchobouillant:
You can use puppeteer-extra-plugin-recaptcha.

berstend:
This is being shown when opening a bug report:
> NOTE: "SiteA.com stopped working" is not a valid stealth-plugin bug report, unless used as an example to show a technical issue while providing specific example code of the problem.
For general purpose usage discussions please use the community discord: http://scraping-chat.cf

Issue #196, opened 2020-05-10
Puppeteer bot detection with out-of-process iframes by user-agent and screen size


ukrexpo:
Real User-Agent is detectable with <iframe> when using "puppeteer-extra-plugin-stealth". To prevent it use "'--disable-features=site-per-process'" argument for puppeteer.launch(). See explanation: https://github.com/puppeteer/puppeteer/issues/2548

Real screen size is also detectable with iframe, to prevent it I used preload() script.

OS: Windows 10
"puppeteer": 3.0.4
"puppeteer-extra": 3.1.9
"puppeteer-extra-plugin-stealth": 2.4.9

Code to reproduce:

``` javascript
const puppeteer = require('puppeteer');
const puppeteerExtraPluginStealth = require('puppeteer-extra-plugin-stealth');
const puppeteerExtraPluginUserAgentOverride = require('puppeteer-extra-plugin-stealth/evasions/user-agent-override');
const {PuppeteerExtra} = require('puppeteer-extra');

function preload(device) {
  Object.defineProperty(navigator, 'platform', {
    value: device.platform,
    writable: true,
  });
  Object.defineProperty(navigator, 'userAgent', {
    value: device.userAgent,
    writable: true,
  });
  Object.defineProperty(screen, 'height', {
    value: device.viewport.height,
    writable: true,
  });
  Object.defineProperty(screen, 'width', {
    value: device.viewport.width,
    writable: true,
  });
  Object.defineProperty(window, 'devicePixelRatio', {
    value: device.viewport.deviceScaleFactor,
    writable: true,
  });
}

const device = {
  userAgent: 'Mozilla/5.0 (Macintosh)',
  viewport: {
    width: 1200,
    height: 800,
    deviceScaleFactor: 1,
    isMobile: false,
    hasTouch: false,
    isLandscape: true,
  },
  locale: 'en-US,en;q=0.9',
  platform: 'Macintosh',
};

(async () => {
  try {
    const pptr = new PuppeteerExtra(puppeteer);
    const pluginStealth = puppeteerExtraPluginStealth();
    pluginStealth.enabledEvasions.delete('user-agent-override'); // Remove this specific stealth plugin from the default set
    pptr.use(pluginStealth);

    const pluginUserAgentOverride = puppeteerExtraPluginUserAgentOverride({
      userAgent: device.userAgent,
      locale: device.locale,
      platform: device.platform,
    });
    pptr.use(pluginUserAgentOverride);

[...]

PR #439, opened 2021-02-25
Add doc for Cross-Origin iFrame

Edweis:
Fixing https://github.com/berstend/puppeteer-extra/issues/437
From https://stackoverflow.com/questions/46529201/puppeteer-how-to-fill-form-that-is-inside-an-iframe


Niek (approved):

berstend:
@Edweis do you know of any other way than to `--disable-web-security`? This is like a nuclear option of death, with many casualties along the way (e.g. any site will be able to fetch your facebook cookies as CORS is disabled). This can also be tested against.

berstend:
Isn't this sufficient?
```
      "--disable-features=IsolateOrigins,site-per-process",
      "--flag-switches-begin --disable-site-isolation-trials --flag-switches-end",
```

berstend:
Related discussion: #196 

Would be great if someone can make a proper test case to confirm failing behavior + fixed behavior when the two launch args are used. Also it'd be worthwhile to document negative side-effects - if they're neglectable we can make an evasion to enable that as the default.

berstend:
Love the followup and further explanation on this, one of the joys of maintaining open-source.

I took it upon myself to properly verify this and low and behold this PR is nonsense:

image
```js
const puppeteer = require("puppeteer")

puppeteer
  .launch({
    headless: false,
    args: [
      "--disable-features=IsolateOrigins,site-per-process,SitePerProcess",
      "--flag-switches-begin --disable-site-isolation-trials --flag-switches-end",
    ],
  })
  .then(async (browser) => {
    await page.goto("chrome://process-internals/")
    await page.waitForTimeout(60 * 1000)
    await browser.close()
  })
```

Thanks!

PR #449, opened 2021-03-11
feat(plugin-stealth): Enhance the navigator.webdriver evasion for Chrome 89.0.4339

### LINKED ISSUE #448: [Bug] stealth evasion `navigator.webdriver` needs to be updated following the Chrome v89 release
The behavior of the `navigator.webdriver` property has been changed in [commit `bfe0108` of Chrome](https://chromium.googlesource.com/chromium/src/+/bfe010892fac3d7731da88478b93ae19953aff1c) like this:

|          | before    | after |
|----------|-----------|-------|
| headful  | undefined | false |
| headless | true      | true  |

Therefore [the `navigator.webdriver` evasion here](https://github.com/berstend/puppeteer-extra/blob/master/packages/puppeteer-extra-plugin-stealth/evasions/navigator.webdriver/index.js) shall be updated accordingly.

This specific commit `bfe0108` has been part of Chrome since major version 89, and recently the stable channel of release has advanced to major version 89, making the fix seemingly more important than before.

starrify:
Fixes #448

### DIFF
--- a/packages/puppeteer-extra-plugin-stealth/evasions/navigator.webdriver/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/navigator.webdriver/index.js
@@ -15,10 +15,16 @@ class Plugin extends PuppeteerExtraPlugin {
     return 'stealth/evasions/navigator.webdriver'
   }
 
-  // Pre Chrome 88.0.4291.0
   async onPageCreated(page) {
     await page.evaluateOnNewDocument(() => {
-      delete Object.getPrototypeOf(navigator).webdriver
+      if (navigator.webdriver === false) {
+        // Post Chrome 89.0.4339.0 and already good
+      } else if (navigator.webdriver === undefined) {
+        // Pre Chrome 89.0.4339.0 and already good
+      } else {
+        // Pre Chrome 88.0.4291.0 and needs patching
+        delete Object.getPrototypeOf(navigator).webdriver
+      }
     })
   }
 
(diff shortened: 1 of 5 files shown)


starrify:
Hmm the CI failure seems irrelevant to my changes. :thinking: 
Locally I have tested only this single evasion.

berstend:
One of the cat&mouse tests fails, probably as they're not updated to handle the new webdriver behavior - checking that now.

berstend:
Ah, it actually says so in the logs: `The following fingerprints failed: [ { name: 'WEBDRIVER', consistent: 1, data: {} } ]`

berstend:
@starrify could you
- add `.filter(check => !(check && check.name === 'WEBDRIVER'))` to [this line](https://github.com/berstend/puppeteer-extra/blob/master/packages/puppeteer-extra-plugin-stealth/test/fpscanner.test.js#L25)
- make `detectionResults` a let and change it afterwards: `Object.fromEntries(Object.entries(detectionResults).filter(([k,v]) => !k.startsWith('navigator.webdriver')))` [here](https://github.com/berstend/puppeteer-extra/blob/master/packages/puppeteer-extra-plugin-stealth/test/cat-and-mouse.test.js#L20)

I can't reproduce this locally even with pptr@7 for some reason but it's clear those tests fail due to new webdriver behavior, it's best just to filter those tests out (we have specific testing anyway). :-)

starrify:
Ah my bad.. should have checked that part of tests as well. Thanks a lot for digging that up!
I'm temporarily away and would come back fixing that later today.

berstend:
Published as `puppeteer-extra-plugin-stealth@2.7.6`
thanks @starrify 👍

PR #456, opened 2021-03-19
convert puppeteer-extra-plugin-stealth to typescript.

UrielCh:
this PR require PR#455. before any merge.


UrielCh:
will resubmit that later

Issue #426, opened 2021-02-12
[Bug] Example typescript code doesn't work - Argument of type '{ headless: boolean; }' is not assignable to parameter of type 'LaunchOptions'.   Object literal may only specify known properties, and 'headless' does not exist in type 'LaunchOptions'.


ajanian:
I am successfully using puppeteer-extra in my typescript project.  I wanted to set headless: false in my launch according to the sample code.  I looked and something appears wrong with my types.  I have stripped the problem down to a trivial sample project.

**Sample Code**

```typescript
import puppeteer from 'puppeteer-extra';
import StealthPlugin from 'puppeteer-extra-plugin-stealth';

puppeteer.use(StealthPlugin());

(async (): Promise<void> => {
    const browser = await puppeteer.launch({ headless: false });
    await browser.close();
})();
```

**Transpile Error**

```bash
❯ tsc -p .
src/index.ts:7:46 - error TS2345: Argument of type '{ headless: boolean; }' is not assignable to parameter of type 'LaunchOptions'.
  Object literal may only specify known properties, and 'headless' does not exist in type 'LaunchOptions'.

7     const browser = await puppeteer.launch({ headless: false });
                                               ~~~~~~~~~~~~~~~


Found 1 error.
```

**Versions**

```bash
❯ npx envinfo@latest --system --binaries --npmPackages '*(puppeteer*|playwright*|automation-extra*|@extra*)'
Need to install the following packages:
  envinfo@latest
Ok to proceed? (y)

  System:
    OS: macOS 11.2
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 54.80 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 15.8.0 - /usr/local/bin/node
    npm: 7.5.0 - /usr/local/bin/npm
  npmPackages:
    puppeteer-extra-plugin-stealth: ^2.7.4 => 2.7.4
```

**Package.json**

```json
{
  "name": "pex",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "Andrew Janian <andrew@janian.net>",
  "license": "ISC",
  "dependencies": {
    "puppeteer-extra-plugin-stealth": "^2.7.4"
  }
}
```

otaviobps:
[...]

PR #461, opened 2021-03-24
full conversion to Typescript.

UrielCh:
Hi,

It's take more time than expected, but all test pass localy.

dependences changes: 
- ava upgraded from 2.4.0 to 3.15.0 (typing issue with 2.4.0)
- puppeteer version is 8.0.0 by default.
- typescript from 3.7.4 to 4.1.5 (needed by typescript 8.0.0)
- @cliqz/adblocker-puppeteer version 1.18.7 to 1.20.3 (needed by typescript 8.0.0)
- ow upgrade version from 0.4.0 to 0.23.0

fixed issue in code
- fix new keyword used on function.
- remove some non exiting parameters
- fix puppeteer-extra-plugin-devtools test (still not passing)
- drop https from https://www.example.com string for local testing. (can be rollback)
- fix invalid  interface PluginData

note:
- one test had been removed form test/REPLSession.test.ts, I do not know how to use ow
- Puppeteer type import had been refactor since my previous PR.
- most of class Plugin had been rename to a proper unique class name.


Xetera:
image

MilanBarande:
I guess this will fix [this issue](https://github.com/berstend/puppeteer-extra/issues/426). @berstend please consider reviewing this one

berstend:
I really appreciate the work @UrielCh put into this, this PR is however too massive (and partially too opinionated) to be merged in as it is.

It's a good reference however and I will use it when cherry picking specific issues.

Most pressing is #426 - I would like to stay backwards compatible with pptr versions < v8 and need to test if there's a way to do this despite the breaking type changes

PR #464, opened 2021-03-27
chore(plugin-stealth): Disable navigator.plugins in chromium 90

berstend:
Starting from chromium 90 the browser will return empty mimetypes and plugin arrays, so we don't want to spoof them then:
https://www.chromestatus.com/features/5741884322349056

v90 is stable in only 2 weeks, so it was necessary to update the stealth plugin to reflect that change.

Luckily the empty special arrays are still present in headless, so we don't even need to spoof those (canary headless):
image

This PR:
- adds `utils.chromiumVersion` to make parsing/comparing semver easy
- disables the `navigator.plugins` evasion in v90 and up

Apologies for the noise caused by prettier :-)

### DIFF
--- a/packages/puppeteer-extra-plugin-stealth/evasions/_utils/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/_utils/index.js
@@ -27,7 +27,7 @@ utils.stripProxyFromErrors = (handler = {}) => {
   // We wrap each trap in the handler in a try/catch and modify the error stack if they throw
   const traps = Object.getOwnPropertyNames(handler)
   traps.forEach(trap => {
-    newHandler[trap] = function () {
+    newHandler[trap] = function() {
       try {
         // Forward the call to the defined proxy handler
         return handler[trap].apply(this, arguments || [])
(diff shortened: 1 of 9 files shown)


berstend:
Jeez, I forgot that `navigator?.userAgent` is a newish feature that fails in older Node versions

Niek (approved):

berstend:
Forgot to add a version check to some tests

sh0cked:
Seems they returned plugins/mime-types arrays back https://bugs.chromium.org/p/chromium/issues/detail?id=1164635

berstend:
@sh0cked thanks for the link 👍 

Damn, I was really looking forward to seeing them gone.
Unfortunately all linked bug reports are private, so we don't even know why they add them back. 😄

berstend:
Converted this to a draft PR until the [situation](https://bugs.chromium.org/p/chromium/issues/detail?id=1164635) shakes out.

momala454:
> 
> 
> @sh0cked thanks for the link 👍
> 
> Damn, I was really looking forward to seeing them gone.
> Unfortunately all linked bug reports are private, so we don't even know why they add them back. 😄

some websites use the presence of PDF plugin on the plugin list to allow to view pdf instead of download them, so this broke the functionnality

berstend:
https://www.chromestatus.com/feature/5741884322349056

> This change was cancelled. See the bug.

_sigh_

Niek:
Looking at https://bugs.chromium.org/p/chromium/issues/detail?id=1164635#c47, it's probably delayed till 91. BTW weird that the referring bugs that let to the reversal are marked private... 🤔

Issue #364, opened 2020-11-01
Solving request interception chain


XavierGeerinck:
Hi Everyone!

Currently I am having the problem that when you have 2 plugins that try to intercept and handle requests, that the dependencies start to fail. (E.g. when I have the adblocker plugin and a personal plugin that both try to block resources, then it starts stating that a request has already been handled).

I agree that this is how it is with Puppeteer and that it will be hard to fix it (sadly enough). But seeing the plugin interface, I think it might be worthwhile to find a solution to this.

Some thoughts that I had:
* Create page hooks (such as the onPageCreated) but then for onPageRequest, onPageRequestFinished, onPageRequestFailed and make them check if a request has been handled already (it should be possible to keep working with requests until an .abort has been called or the request finished?)
* Add a weight system to allow certain plugins to come before others (e.g. adblocker to block requests before other plugins).

I'd love to hear some thoughts from the community to see if this is feasible or not? Looking forward to the responses!

berstend:
@XavierGeerinck thanks for the well thought-out comment 👍

It's been a while since I worked on that aspect, but I remember the main annoyance being that when request interception is enabled someone needs to decide if the request shall be continued or aborted (but that decision can only happen once).

That makes it a bit annoying to work with from a plugin framework perspective 😄 If I remember correctly puppeteer is holding an internal state for this.

If all code happens within plugins this would be relatively simple to fix, it get's more tricky when we need to take care of code in the user's scripts as well (by patching `.enableRequestInterception`) and the like.

I would need to refresh my memory on the issues I ran into while working on a solution to this a while back, happy for other input on this though.

XavierGeerinck:
Well I am personally fine if plugins have to be written for this to work inside the core, but I also believe a better solution could be found. The solution I personally was thinking was to indeed handle this within the plugin framework and:
* add the appropriate hooks, then when they exist make the hooks return abort or continue and finally decide in the framework to continue or not :)
* Add a specific flag to show what was decided (so that in the code you can check if the request was indeed handled or not)

Is this something that could be added? Would love to see this upstream!

In the meantime I also opened a request on puppeteer for official plugin support as well as solving this use case: [https://github.com/puppeteer/puppeteer/issues/6567](https://github.com/puppeteer/puppeteer/issues/6567)

let me know what you think!

knuppe:
A simple solution would be.
```javascript
page.on('request', async (req) => {
    if (req._interceptionHandled) return;
    /* ... */
}

[...]

Issue #470, opened 2021-04-05
[Bug] unhandledRejection (Target.attachToTarget)


wintrdotcom:
**Describe the bug**

Hey there, reproducing my code snippet will lead to an `unhandledRejection`.

It basically opens a Browser instance and closes it before creating a new page, therefore, it fails as expected. The problem is that the error is not correctly catched.

If I reproduce the same code snippet commenting `puppeteer.use(pluginStealth())`, the error will be correctly catched so I guess the problem comes from `puppeteer-extra` rather than from `puppeteer` itself.

**Code Snippet**

```javascript
const puppeteer = require('puppeteer-extra')
const pluginStealth = require('puppeteer-extra-plugin-stealth')

puppeteer.use(pluginStealth())

puppeteer.launch({
	headless: false,
	args: [
		'--window-position=0,0',
		'--disable-infobars',
		'--disable-extensions',
		'--disable-accelerated-2d-canvas',
		'--disable-gpu',
		'--window-size=1920x1080',
		'--disable-dev-shm-usage',
		'--disable-setuid-sandbox',
		'--no-first-run',
		'--no-sandbox',
		'--no-zygote',
		'--single-process',
		'--deterministic-fetch',
		'--enable-features=NetworkService',
		'--ignore-certificate-errors'
	]
})
.then((browser) => {
    browser.close()

	return browser.newPage()
})
.catch((err) => {
    console.log(err)
})
```

**What is happening**

```javascript
(node:18424) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:18424) UnhandledPromiseRejectionWarning: Error: Protocol error (Target.attachToTarget): No target with given id found
    at C:\Users\xxxx\Desktop\lol\node_modules\puppeteer\lib\cjs\puppeteer\common\Connection.js:71:63
    at new Promise (<anonymous>)
    at Connection.send (C:\Users\xxxx\Desktop\lol\node_modules\puppeteer\lib\cjs\puppeteer\common\Connection.js:70:16)
    at Connection.createSession (C:\Users\xxxx\Desktop\lol\node_modules\puppeteer\lib\cjs\puppeteer\common\Connection.js:141:42)
    at Target._sessionFactory (C:\Users\xxxx\Desktop\lol\node_modules\puppeteer\lib\cjs\puppeteer\common\Browser.js:178:91)
[...]

PR #484, opened 2021-04-23
feat(plugin-stealth): Catch and mute certain Protocol errors

benww:
We recently started using [`enchant-puppeteer`](https://github.com/berstend/puppeteer-extra/issues/364#issuecomment-757049890) and doing our own asynchronous request interception while also using `puppeteer-extra-plugin-stealth`.  It seems like this combination makes it much more likely for this error to happen, especially on pages with redirects. The error was causing the whole process to be killed.

This seems relevant: https://github.com/puppeteer/puppeteer/issues/2258#issuecomment-380647459

We aren’t sure this is the right way to handle this problem. However, this seems like an improvement.

### DIFF
--- a/packages/puppeteer-extra-plugin-stealth/evasions/sourceurl/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/sourceurl/index.js
@@ -27,7 +27,18 @@ class Plugin extends PuppeteerExtraPlugin {
     page._client.send = (function(originalMethod, context) {
       return async function() {
         const [method, paramArgs] = arguments || []
-        const next = () => originalMethod.apply(context, [method, paramArgs])
+        const next = async () => {
+          try {
+            return await originalMethod.apply(context, [method, paramArgs])
+          } catch(error) {
+            // This seems to happen sometimes when redirects cause other outstanding requests to be cut short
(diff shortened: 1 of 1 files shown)


berstend:
@benww did you consider using a global `unhandledRejection` handler to capture/mute those specific errors? https://github.com/berstend/puppeteer-extra/issues/470#issuecomment-815102751

I'm not 100% sure we'd want to silence errors like this in the framework automatically. Maybe adding a opt-in flag and a possible list of errors to silence would be an option 🤔

benww:
@berstend I could use a global `unhandledRejection` handler. Maybe I'm understanding the error wrong, but it seemed to me that this error is irrelevant to people using the plugin because it means the network request no longer exists and there was no failure to make the request "stealthy".

berstend:
@benww Ah, I overlooked that your PR is only affecting the `sourceurl` evasion (we had similar discussions before with PRs to mute Protocol errors globally).

I guess we can merge that in then, though it might make sense to thing about configuring handling of Protocol errors globally in the future (there's others, like `Target closed`, etc). :-)

berstend (approved):

berstend:
```
Successfully published:
 - puppeteer-extra-plugin-stealth@2.7.8
```
thanks @benww 👍

benww:
@berstend Thank you! I appreciate it. We have only encountered that protocol error so far, but we will keep that in mind going forward.

PR #486, opened 2021-04-24
feat(plugin-recaptcha): Add useEnterpriseFlag and useActionValue options

berstend:
This PR is based on [previous discussions](https://github.com/berstend/puppeteer-extra/pull/400#issuecomment-823183394).

The main idea is to pass `enterprise=1` to 2captcha when an enterprise recaptcha is being detected. According to @divinedeveloper using the enterprise flag improved his success rate on hulu.com.

Unfortunately in my own testing (using my own [test sites](https://github.com/berstend/berstend.github.io/tree/main/static/recaptcha)) setting the enterprise flag made things worse (`ERROR_CAPTCHA_UNSOLVABLE`) so I'm keeping this opt-in for now, until we fully understood why this flag is not always improving the solving success rate.

In addition I added support for passing the optional `action` parameter (v3/enterprise) to 2captcha, this is opt-out as it didn't make a negative difference in my testing.

### Usage

```js
  const recaptchaPlugin = RecaptchaPlugin({
    provider: {
      id: '2captcha',
      token: process.env.TWOCAPTCHA_TOKEN,
      opts: {
        useEnterpriseFlag: false, // default: false
		useActionValue: true // default: true
      }
    }
  })
  puppeteer.use(recaptchaPlugin)
```

(Apologies for the linting noise)

### DIFF
--- a/packages/puppeteer-extra-plugin-recaptcha/src/content.ts
+++ b/packages/puppeteer-extra-plugin-recaptcha/src/content.ts
@@ -1,11 +1,11 @@
 import * as types from './types'
 
 export const ContentScriptDefaultOpts: types.ContentScriptOpts = {
-  visualFeedback: true,
+  visualFeedback: true
 }
 
 export const ContentScriptDefaultData: types.ContentScriptData = {
-  solutions: [],
+  solutions: []
 }
 
 /**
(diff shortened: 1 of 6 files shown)


berstend:
```
Successfully published:
 - puppeteer-extra-plugin-recaptcha@3.3.9
```

PR #488, opened 2021-04-25
feat(plugin-recaptcha): Support more recaptcha domain combos

berstend:
Based on [this discussion](https://github.com/berstend/puppeteer-extra/pull/482#issuecomment-826281337) and a follow up to #482.

At the end of the day there are more possible URLs for the recaptcha scripts to originate from (with/without www, using http or https, etc). 

This PR refactors the selector stuff to use a generator, which means we now support a grand total (and hopefully final) number of 16 recaptcha domains 😄 

image


berstend:
```
Successfully published:
 - puppeteer-extra-plugin-recaptcha@3.3.10
```

PR #501, opened 2021-05-18
Support Yarn V2 PnP

richardfuca:
This PR adds support for Yarn V2's PnP.  Also adds some missing dependencies that Yarn V2 would complain about.

Full changes:
* Add Yarn V2 PnP support
* Add missing peer dependencies that Yarn would complain about
* [plugin-adblocker] Move puppeteers types into dependencies for fix TS typing issue

### DIFF
--- a/packages/puppeteer-extra/src/index.ts
+++ b/packages/puppeteer-extra/src/index.ts
@@ -365,7 +365,7 @@ export class PuppeteerExtra implements VanillaPuppeteer {
       let dep = null
 
       const pnpInUse = process.versions.pnp != null;
-      const yarnExtensionsInstalled = require('pnpapi').VERSIONS.getAllLocators != null;
+      const yarnExtensionsInstalled = pnpInUse && require('pnpapi').VERSIONS.getAllLocators != null;
 
       try {
         // Try to require and instantiate the stated dependency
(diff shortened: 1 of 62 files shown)


packages/puppeteer-extra/src/index.ts
```diff
@@ -384,6 +406,27 @@ export class PuppeteerExtra implements VanillaPuppeteer {
...
   }
 
+  /**
+   * Resolves a plugin through Yarn's PnP system.
+   *
+   * Requires Yarn v2.3.0+ due to using the Yarn extension method getAllLocators.
+   *
+   * @private
+   */
+  private resolvePnpPlugin(pluginName: string, requiredBy: string) {
```
merceyz, line 416: Wont this be a bit problematic if you have multiple versions of `requiredBy`? It would be safer to have `requiredBy` be a file path

richardfuca: Good point, I've changed it to use paths instead. Though I don't really like how I've implemented it, any ideas on how to improve it?

merceyz: Well you can get rid of the PnP specific logic and use
```js
require(require.resolve(pluginName, { paths: [requiredBy] }))
```

Just running a `require(pluginName)` is a problem in `node_modules` as well so if anything this is a bugfix rather than adding PnP support
https://yarnpkg.com/advanced/rulebook#packages-should-only-ever-require-what-they-formally-list-in-their-dependencies

richardfuca: Ah, that works much better. Thanks!

Issue #421, opened 2021-02-04
[Bug] My software stucks on browser.close with plugin-stealth 2.7.x version


Dam998:
My software stucks on browser.close with last version (puppeteer-extra-plugin-stealth)

I'm using more than 1 browser open at the same time, and when I'm trying to close 1 my software stucks.
this has never happened with version 2.6.7 or before

my code:

```javascript
var pages = await this.browser.pages()
for (let i = 0; i < pages.length; i++) {
    await pages[i].close();
}
await this.browser.close();
```

Tested with 2.7.x version

if i downgrade to 2.6.7 everything works fine

berstend:
How are you opening the browsers, `.launch` or `.connect`?

Dam998:
.launch

berstend:
Could you remove the `user-agent-override` evasion and test if it works then?

```js
const pluginStealth = require('puppeteer-extra-plugin-stealth')()
pluginStealth.enabledEvasions.delete('user-agent-override')
puppeteer.use(pluginStealth)
```

berstend:
The issue might be related to https://github.com/berstend/puppeteer-extra/pull/413 - we save some state there and added functionality depending on how the browser is launched

Dam998:
if I remove user-agent-override is working

can I ask you why? if this is correct and nothing change at the end
or if you suggest me to launch browser in a different way

berstend:
So it works without the `user-agent-override` evasion?

The reason is to narrow down the problem, we made a larger update to the evasion so it makes sense the issue is somewhere there.

If removing `user-agent-override` fixes the issue in 2.7.x then you just need to wait for us to develop a fix, until then you need to use the older version for your use-case.

berstend:
@Dam998 did you mean opening multiple browsers or multiple pages in a single browser?

Your words indicate the former, your code the latter

Dam998:
opening multiple browsers (more browsers, each browser with 2 pages)

I created a class that on start() open a browser and then close it at the end, this is why

Niek:
@Dam998 do you also mind sharing a minimal reproducible code sample so we can check?

berstend:
Are you able to make a small self-contained test file? this helps us reproduce the issue quickly. Thanks!

Dam998:
I'll try, give me some minutes

Dam998:
I reproduced it with this simple script:

```javascript
const puppeteer = require('puppeteer-extra');
[...]

PR #502, opened 2021-05-19
fs.rmdirSync() deprecated, use rmSync()

An-GG:
We get this warning from **puppeteer-extra-plugin-user-data-dir**:
```bash
(node:52638) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead (Use `node --trace-deprecation ...` to show where the warning was created)
```

Swaps `fse.rmdirSync()` with `fse.rmSync()`, which has identical behavior (both throw error when dir does not exist, both recursively remove subdirectories and files), successfully eliminates warning.

### DIFF
--- a/packages/puppeteer-extra-plugin-user-data-dir/index.js
+++ b/packages/puppeteer-extra-plugin-user-data-dir/index.js
@@ -69,7 +69,7 @@ class Plugin extends PuppeteerExtraPlugin {
     try {
       // We're doing it sync to improve chances to cleanup
       // correctly in the event of ultimate disaster.
-      fse.rmdirSync(this._userDataDir, { recursive: true })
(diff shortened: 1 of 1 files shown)


Niek (approved):

Maxim-Mazurok:
LGTM, test failure seems unrelated, I'd rebase/re-run it.

berstend:
Apparently that won't fix the issues some windows users are experiencing:
https://github.com/berstend/puppeteer-extra/issues/421#issuecomment-917930695

```js
fs.rm(this._userDataDir, { recursive: true, force: true }, console.log)
```
works on the other hand - closing this PR in favor of a new one with an async method. :)

PR #506, opened 2021-05-26
fix(plugin-stealth): Make makeHandler consistent with real getters

anarcal:
Fixes an inconsistency when an overridden getter is called via call or apply with more than zero arguments.
```js
const obj = { get foo() { return 1 } };
utils.replaceGetterWithProxy(obj, "foo", utils.makeHandler().getterValue(2));

console.log(obj.foo); // 2
console.log(Object.getOwnPropertyDescriptor(obj, "foo").get.call(obj, null)); // 1
```

Checking if `args` is truthy is redundant since it is guaranteed to be an array-like.

### DIFF
--- a/packages/puppeteer-extra-plugin-stealth/evasions/_utils/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/_utils/index.js
@@ -489,7 +489,7 @@ utils.makeHandler = () => ({
     apply(target, ctx, args) {
       // Let's fetch the value first, to trigger and escalate potential errors
       // Illegal invocations like `navigator.__proto__.vendor` will throw here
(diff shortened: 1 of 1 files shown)


berstend (approved):
LGTM

berstend:
```
Successfully published:
 - puppeteer-extra-plugin-anonymize-ua@2.3.1
 - puppeteer-extra-plugin-devtools@2.3.1
 - puppeteer-extra-plugin-stealth@2.8.1
 - puppeteer-extra-plugin-user-data-dir@2.3.1
 - puppeteer-extra-plugin-user-preferences@2.3.1
 - puppeteer-extra@3.2.1
```

PR #515, opened 2021-06-10
Send options to click method

merloisac:
Update to send options to new humanize click method


berstend:
Closing for now, that plugin is in beta mode :)

PR #537, opened 2021-08-12
Fix launch options

lexagr:
https://github.com/berstend/puppeteer-extra/issues/426


berstend:
Thanks for the PR @lexagr - it was easier for me to fix this through #555 👍

Issue #532, opened 2021-07-25
CloudFlare update?


ElijahDD:
Hi. CloudFlare update their JS Challenge? Stealth plugin can't bypass it now.

itsaphel:
Confirm, but works with `headless: false`. It actually doesn't start the JS challenge at all but just goes to `Access denied`, which makes me think this is some sort of TLS fingerprinting at work, but just a guess.

prescience-data:
![image](https://user-images.githubusercontent.com/65471523/127300455-814d122a-8ada-48b0-920f-925e0cec3f88.png)

taewookim:
[Related](https://github.com/berstend/puppeteer-extra/issues/236)

regseb:
Cloudflare doesn't work, it redirects in loop on the page:

> Checking your browser before accessing XXX.
> This process is automatic. Your browser will redirect to your requested content shortly.
>
> Please allow up to 5 seconds…
>
> DDoS protection by Cloudflare

The following error is displayed in the console.

```
[[[ERROR]]]: Message: Uncaught TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Window'
    --- property 'window' closes the circle - URL:  - Line: 5 - Column: 14655 - Error object: {}
{i: "F0qg3w2C0uk=", h: "6e72b2ba904e2ec", tH: "", t: -1630431698613}h: "6e72b2ba904e2ec"i: "F0qg3w2C0uk="t: -1630431698613tH: ""[[Prototype]]: Objectconstructor: ƒ Object()hasOwnProperty: ƒ hasOwnProperty()isPrototypeOf: ƒ isPrototypeOf()propertyIsEnumerable: ƒ propertyIsEnumerable()toLocaleString: ƒ toLocaleString()toString: ƒ toString()valueOf: ƒ valueOf()__defineGetter__: ƒ __defineGetter__()__defineSetter__: ƒ __defineSetter__()__lookupGetter__: ƒ __lookupGetter__()__lookupSetter__: ƒ __lookupSetter__()get __proto__: ƒ __proto__()set __proto__: ƒ __proto__()

Uncaught TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Window'
    --- property 'window' closes the circle
    at JSON.stringify (<anonymous>)
    at eval (eval at C.<computed> (v1?ray=68780cf65e342209:formatted:2161), <anonymous>:5:14655)
[...]

PR #545, opened 2021-09-03
fix(plugin-stealth): Improve iframe.contentWindow evasion

regseb:
I think I found how Cloudflare detects puppeteer: `iframe.contentWindow[0]` returns `undefined` in Chrome and returns `window` with puppeteer (and _iframe.contentWindow_ evasion). I modify `contentWindowProxy` to returns `undefined`.

Here is a case to see the difference between Chrome and puppeteer (with _iframe.contentWindow_ evasion).

```HTML
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
  </head>
  <body>
    <script>
        // https://github.com/berstend/puppeteer-extra/blob/master/packages/puppeteer-extra-plugin-stealth/evasions/iframe.contentWindow/index.js
        const contentWindowProxy = {
            get(target, key) {
                // Now to the interesting part:
                // We actually make this thing behave like a regular iframe window,
                // by intercepting calls to e.g. `.self` and redirect it to the correct thing. :)
                // That makes it possible for these assertions to be correct:
                // iframe.contentWindow.self === window.top // must be false
                if (key === 'self') {
                    return this
                }
                // iframe.contentWindow.frameElement === iframe // must be true
                if (key === 'frameElement') {
                    return iframe
                }
                return Reflect.get(target, key)
            }
        }

        const body = document.querySelector('body')
        const iframe = document.createElement('iframe')

        const proxy = new Proxy(window, contentWindowProxy)
        Object.defineProperty(iframe, 'contentWindowStealth', {
            get() {
                return proxy
            },
            set(newValue) {
                return newValue; // contentWindow is immutable
            },
            enumerable: true,
            configurable: false
        })

        iframe.srcdoc = 'foobar'
        console.log("proxy before append iframe", proxy);
        body.appendChild(iframe)
        console.log("proxy after append iframe", proxy);

        console.log("iframe.contentWindow[0]", iframe.contentWindow[0])
        console.log("iframe.contentWindowStealth[0]", iframe.contentWindowStealth[0])
    </script>
  </body>
</html>
```

With this pull request, I no longer have the error indicated in the issue https://github.com/berstend/puppeteer-extra/issues/532#issuecomment-909472096, but puppeteer still can't pass the Cloudflare redirect.

### DIFF
--- a/packages/puppeteer-extra-plugin-stealth/evasions/iframe.contentWindow/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/iframe.contentWindow/index.js
@@ -43,6 +43,10 @@ class Plugin extends PuppeteerExtraPlugin {
               if (key === 'frameElement') {
                 return iframe
               }
+              // Intercept iframe.contentWindow[0] to hide the property 0 added by the proxy.
+              if (key === '0') {
+                  return undefined
+              }
               return Reflect.get(target, key)
             }
           }
--- a/packages/puppeteer-extra-plugin-stealth/evasions/iframe.contentWindow/index.test.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/iframe.contentWindow/index.test.js
@@ -51,6 +51,42 @@ test('stealth: will not break iframes', async t => {
   t.is(realReturn, 'TESTSTRING')
 })
 
+test('vanilla: will not have contentWindow[0]', async t => {
+  const browser = await vanillaPuppeteer.launch({ headless: true })
+  const page = await browser.newPage()
+
+  const zero = await page.evaluate(returnValue => {
+    const { document } = window // eslint-disable-line
+    const body = document.querySelector('body')
+    const iframe = document.createElement('iframe')
+    iframe.srcdoc = 'foobar'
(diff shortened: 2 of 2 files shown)


kasjp (approved):

mfunder:
@regseb 
This solves the problem for me. Perhaps this pull request would get more attention, and be easier for others to find and apply, if the title included "(CloudFlare fix)"?

berstend (approved):
Looking good, thanks for adding the tests @regseb 👍

berstend:
```
Successfully published:
 - puppeteer-extra-plugin-stealth@2.7.10
```
thanks @regseb 👍 

cc @pussyslain @mfunder

norstbox:
@mfunder
So, you could pass CloudFlare JS challenge with this patch? Do you use puppeteer in `headless: false` mode?

mfunder:
@norstbox It seems to only work with headful: true.
It also seems that Cloudflare is running several anti-bot scripts, and this patch only helps with one of them. But since Cloudflare uses a different script on each page refresh, and refreshes by itself on failure, it eventually gets through.

Issue #546, opened 2021-09-09
[Bug] Code breaks when tab no longer exists on pdf download


bubbomb:
**Describe the bug**

We use puppeteer to connect to a remote instance of chrome that has these user preferences:

```
{
"plugins": {
      "always_open_pdf_externally": false,
      "plugins_disabled": [
        "Chrome PDF Viewer"
      ],
      "plugins_list": []
    }
...
"download": {
      "default_directory": "/home/node/Downloads",
      "directory_upgrade": true,
      "extensions_to_open": "applications/pdf",
      "prompt_for_download": false
    },
...
}
```

When opening a pdf (or other download link) the puppeteer-extra code creates an uncaught exception. It appears that the code is checking for a new tab. However, the user settings don't open a new tab when downloading a pdf file, so the code breaks.

I expected puppeteer-extra to be able to handle this exception.

**Code Snippet**

```javascript
const puppeteer = require('puppeteer-extra')

;(async () => {
  const connectionString = 'asdf' // chrome connection string with the above user preferences

  const browser = await puppeteer.connect(connectionString)
  const page = await browser.newPage()
  
  const downloadLink = 'www.site-with-download.com/download' // change to an actual download link
  await page.goto(downloadLink)
})()
```

**Versions**

  System:
    OS: macOS 11.5.2
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 82.44 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.16.1 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 6.14.12 - /usr/local/bin/npm

bubbomb:
The PR seems to solve the issue that I've been seeing. Please let me know what changes you'd like to see.

berstend:
@bubbomb I guess this might be fixed by #555 - please ping if that's not the case

bubbomb:
> @bubbomb I guess this might be fixed by #555 - please ping if that's not the case

Yes this works great. Thanks for the updated!

PR #547, opened 2021-09-09
fix: tab closed error

bubbomb:
FIxes an error that occurs when a downloaded pdf doesn't open a new tab. See issue #546


berstend:
Thanks for the PR @bubbomb - it was easier to do it while working on #555.

bubbomb:
> Thanks for the PR @bubbomb - it was easier to do it while working on #555.

No problem glad you were able to get it released with #555. It fixed our issues on our end!

Issue #556, opened 2021-09-18
Deprecation warning on rmDirSync


ppassmannpriv:
https://github.com/berstend/puppeteer-extra/blob/713f9fa5ee2cc4656ee992ce8e4761cb31c3699e/packages/puppeteer-extra-plugin-user-data-dir/index.js#L70

This should become rmSync

https://nodejs.org/api/deprecations.html#DEP0147

PR #557, opened 2021-09-18
[FIX][berstend/puppeteer-extra#556] Change to new function according to docs.

ppassmannpriv:
Just a small change, because on newer node.js versions rmDirSync etc. is deprecated.

Resolves berstend/puppeteer-extra#556


berstend:
Thanks @ppassmannpriv - closing in favor of #564 for backwards compat

PR #558, opened 2021-09-19
fix(recaptcha) findRecaptcha() does not wait indefinitely to timeout

tanguc:
findRecaptcha() does not wait indefinitely to timeout when not finding captchas

### DIFF
--- a/packages/puppeteer-extra-plugin-recaptcha/src/index.ts
+++ b/packages/puppeteer-extra-plugin-recaptcha/src/index.ts
@@ -133,7 +133,7 @@ export class PuppeteerExtraPluginRecaptcha extends PuppeteerExtraPlugin {
         (function() {
           return Object.keys((window.___grecaptcha_cfg || {}).clients || {}).length
         })()
-      `,
+      `, '',
           { polling: 200, timeout: 10 * 1000 }
         )
         .catch(this.debug)
@@ -150,7 +150,7 @@ export class PuppeteerExtraPluginRecaptcha extends PuppeteerExtraPlugin {
         (function() {
           return window.hcaptcha
         })()
-      `,
+      `, '',
         { polling: 200, timeout: 10 * 1000 }
       )
       this.debug('wait:hasHcaptchaScriptTag - end', new Date()) // used as timer


berstend (changes requested):
@tanguc can you provide more context?

Your function signature doesn't match with the API:
https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#pagewaitforfunctionpagefunction-options-args

tanguc:
> 
> 
> @tanguc can you provide more context?
> 
> Your function signature doesn't match with the API:
> https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#pagewaitforfunctionpagefunction-options-args

I did see the `Page` type was coming from Puppeteer's one, on my side I was talking about https://playwright.dev/docs/api/class-page#page-wait-for-function

The second arg is the arg(s) which will be given to the first arg (function)

Now, how should we deal with that now ?

berstend:
@tanguc can you describe the problem you're trying to solve? The code you modified is just convenience in case the developer immediately called `page.findRecaptcha()`, before the page has fully loaded yet.

If you're experiences timeout related issues you could also just wait for the page to be fully loaded (or even the captcha iframe to appear in the DOM) before calling the recaptcha plugin?

berstend:
> findRecaptcha() does not wait indefinitely to timeout when not finding captchas

We don't want to wait indefinitely, as otherwise calling `page.solveRecaptchas()` on a page without captchas would never return

tanguc:
> 
> 
> @tanguc can you describe the problem you're trying to solve? The code you modified is just convenience in case the developer immediately called `page.findRecaptcha()`, before the page has fully loaded yet.
> 
> If you're experiences timeout related issues you could also just wait for the page to be fully loaded (or even the captcha iframe to appear in the DOM) before calling the recaptcha plugin?

I'm already calling `page.findRecaptcha()` only after I did wait the network to idle 
However in the case where the page do not contain any captcha it will indefinitely wait, and in my case I'm using playwright behind as well as the `Page`, the playwright's signature `https://playwright.dev/docs/api/class-page#page-wait-for-function` one is completely different from the Puppeteer's one

no4nou:
при использовании Xevil дает ошибку Error: 2captcha error: XEVIL INTERNAL ERROR: PROBABLY, INCORRECT REQUEST

Issue #543, opened 2021-08-30
[Bug] Epic Games using iframe.contentWindow evasion recursion as a detection


claabs:
**Describe the bug**

I'm working on an automation for epicgames.com. They use a home-baked automation detection system called Talon. It's very obfuscated, but I've been able to discover how they were detecting my automations.

Epic Games generates a fingerprint with many properties using browser APIs. One them called `sd_recurse` was recently added. It is set to `true` on puppeteer with stealth plugin, and `false` on browsers.

**Code Snippet**

Here's the code Epic Games runs to populate this variable:

```js
var iframe = document.createElement("iframe")
try {
    return !!iframe.srcdoc && '' !== iframe.srcdoc
} catch (err) {
    return true;
}
```

In normal browsers, `iframe.srcdoc` resolves to `""`.

However, In puppeteer running the stealth plugin, `iframe.srcdoc` throws the error:
```
RangeError: Maximum call stack size exceeded
    at HTMLIFrameElement.get (eval at <anonymous> (epic:7), <anonymous>:51:15)
    at HTMLIFrameElement.get (eval at <anonymous> (epic:7), <anonymous>:51:30)
    at HTMLIFrameElement.get (eval at <anonymous> (epic:7), <anonymous>:51:30)
    at HTMLIFrameElement.get (eval at <anonymous> (epic:7), <anonymous>:51:30)
    at HTMLIFrameElement.get (eval at <anonymous> (epic:7), <anonymous>:51:30)
    at HTMLIFrameElement.get (eval at <anonymous> (epic:7), <anonymous>:51:30)
    at HTMLIFrameElement.get (eval at <anonymous> (epic:7), <anonymous>:51:30)
    at HTMLIFrameElement.get (eval at <anonymous> (epic:7), <anonymous>:51:30)
    at HTMLIFrameElement.get (eval at <anonymous> (epic:7), <anonymous>:51:30)
    at HTMLIFrameElement.get (eval at <anonymous> (epic:7), <anonymous>:51:30)
```

Luckily, if I disable the `iframe.contentWindow` evasion, the automation is no longer detected:
```js
const stealth = StealthPlugin();
stealth.enabledEvasions.delete('iframe.contentWindow');
puppeteer.use(stealth);
```

Somehow, the `iframe.contentWindow` evasion is causing some kind of recursion on `srcdoc` which Epic Games is using to detect the stealth plugin.

**Versions**

System:
    OS: Windows 10 10.0.19043
[...]

PR #562, opened 2021-09-23
fix(plugin-stealth): Improve iframe.srcdoc evasion

regseb:
Thanks to the good analysis in the issue https://github.com/berstend/puppeteer-extra/issues/543, I fixed an infinite recursion problem when calling `iframe.srcdoc` without assigning it a value.


berstend (approved):
Looking good, thanks for adding tests @regseb 👍

berstend:
```
Successfully published:
 - puppeteer-extra-plugin-anonymize-ua@2.3.1
 - puppeteer-extra-plugin-devtools@2.3.1
 - puppeteer-extra-plugin-stealth@2.8.1
 - puppeteer-extra-plugin-user-data-dir@2.3.1
 - puppeteer-extra-plugin-user-preferences@2.3.1
 - puppeteer-extra@3.2.1
```

Issue #522, opened 2021-06-22
[Bug] Allow websites to assign to `document.createElement`


dkthehuman:
**Describe the bug**

`puppeteer-extra-plugin-stealth` causes an error when loading https://web.whatsapp.com due to the [`iframe.contentWindow` evasion](https://github.com/berstend/puppeteer-extra/tree/puppeteer-extra@3.1.18/packages/puppeteer-extra-plugin-stealth/evasions/iframe.contentWindow).

WhatsApp tries to monkey-patch `document.createElement` but it's unable to due to the `document.createElement` property having `writable` set to `false` from the following lines in the evasion:
- https://github.com/berstend/puppeteer-extra/blob/puppeteer-extra%403.1.18/packages/puppeteer-extra-plugin-stealth/evasions/iframe.contentWindow/
- https://github.com/berstend/puppeteer-extra/blob/adcc26325c9a01febcd152e5fb79555b29187a5f/packages/puppeteer-extra-plugin-stealth/evasions/_utils/index.js#L309

As a result, the page does not load and there's an error in the console: `vendor1~bootstrap_qr.cbfe7781bf47cd2c3488.js:34 TypeError: Cannot assign to read only property 'createElement' of object '#<HTMLDocument>'`

**Code Snippet**

```javascript
const chromium = require("chrome-aws-lambda");
const { addExtra } = require("puppeteer-extra");
const StealthPlugin = require("puppeteer-extra-plugin-stealth");

(async () => {
  const puppeteerExtra = addExtra(chromium.puppeteer);
  puppeteerExtra.use(
    StealthPlugin({
      enabledEvasions: new Set([
        "chrome.app",
        "chrome.csi",
        "chrome.loadTimes",
        "chrome.runtime",
        // Remove following line to fix issue
        "iframe.contentWindow",
        "media.codecs",
        "navigator.hardwareConcurrency",
        "navigator.languages",
        "navigator.permissions",
        "navigator.plugins",
        "navigator.webdriver",
        "sourceurl",
        "user-agent-override",
        "webgl.vendor",
        "window.outerdimensions",
      ]),
    })
  );
  const browser = await puppeteerExtra.launch({
    args: chromium.args,
    defaultViewport: chromium.defaultViewport,
    ignoreHTTPSErrors: true,
    headless: false,
  });
  const page = await browser.newPage();

  await page.goto("https://web.whatsapp.com");

  // await browser.close();
})();
```

**Versions**

```
System:
[...]

PR #565, opened 2021-09-23
fix(plugin-stealth): allow to re-define property

regseb:
To fix the issue https://github.com/berstend/puppeteer-extra/issues/522 and prevent websites from detecting puppeteer when trying to redefine properties.


packages/puppeteer-extra-plugin-stealth/evasions/_utils/index.js
```diff
@@ -306,7 +306,7 @@ utils.replaceWithProxy = (obj, propName, handler) => {
   const originalObj = obj[propName]
   const proxyObj = new Proxy(obj[propName], utils.stripProxyFromErrors(handler))
 
-  utils.replaceProperty(obj, propName, { value: proxyObj })
+  utils.replaceProperty(obj, propName, { configurable: true, value: proxyObj })
```
berstend, line 309: I'm not sure that's needed 🤔 from my understanding of the spec: descriptors that are not defined in `defineProperty` will simply be kept with their original value, also things that had `configurable: false` before can't be modified by us - haven't tested/confirmed the original issue yet though

regseb: I can add the `descriptorOverrides` parameter to let the code calling `replaceWithProxy()` choose.

```JavaScript
utils.replaceWithProxy = (obj, propName, handler, descriptorOverrides = {}) => {
  const originalObj = obj[propName]
  const proxyObj = new Proxy(obj[propName], utils.stripProxyFromErrors(handler))

  utils.replaceProperty(obj, propName, { ...descriptorOverrides, value: proxyObj })
  utils.redirectToString(proxyObj, originalObj)

  return true
}
```

regseb:
@berstend I refactored the _iframe.contentWindow_ evasion to only override the `iframe.contentWindow` property. This pull request fixes the problem of:

- https://github.com/berstend/puppeteer-extra/issues/522;
- `Object.getOwnPropertyDescriptor(iframe, 'contentWindow')` which doesn't return `undefined`;
- `iframe.contentWindow` which doesn't return `null` when `iframe.srcdoc` is defined and the `iframe` isn't attached in the DOM.

MScheisse:
Something is still not right Headless && Headful caught with KASADA Nike whenever using stealth plugin

Whenever I load https://www.nike.com/launch using extra-stealth plugin, kasada `/tl` endpoint that is supposed to return `"reload":true` is not

I am pretty sure it has something to do with evasions iframe.content window

MScheisse:
> Something is still not right Headless && Headful caught with KASADA Nike whenever using stealth plugin
> 
> Whenever I load https://www.nike.com/launch using extra-stealth plugin, kasada `/tl` endpoint that is supposed to return `"reload":true` is not
> 
> I am pretty sure it has something to do with evasions iframe.content window

@regseb I am also able to validate, that even with bypassing kasada solution, something with the iframe manipulations happening is not right, and ruining the application from running as it should + also invalidating my current valid kasada cookies

Please lmk your thoughts upon this matter

berstend:
LGTM, definitely an improvement 👍

berstend:
```
Successfully published:
 - puppeteer-extra-plugin-stealth@2.10.0
 ```

berstend:
@regseb unfortunately we got reports that this will break captcha boxes, e.g.
https://democaptcha.com/demo-form-eng/hcaptcha.html

with the new `iframe.contentWindow` evasion clicking the checkbox will result in a forever loading icon in the hcaptcha checkbox frame

berstend:
Reversal PR with more tests around captcha iframes can be found in #657

PR #569, opened 2021-09-28
fix(plugin-devtools): Add localtunnel error event

abowcut:
As mentioned on many [issues](https://github.com/localtunnel/localtunnel/issues/258) on the localtunnel [github](https://github.com/localtunnel/localtunnel), using the public localtunnel server leads to crashes quite frequently as the public server becomes overloaded many times throughout the day.

Currently, when using the devtools plugin, if the public localtunnel server is having one of its "episodes" and the connection refused error is thrown the entire node process that puppeteer is running in crashes.

This pull request simply adds a listener to the tunnel for the error event and handles it so that even though the cdp might not be working for that browser at least the node process can continue to run.
To me this is definitely preferable over having the whole node process running puppeteer exit.

### DIFF
--- a/packages/puppeteer-extra-plugin-devtools/lib/RemoteDevTools.js
+++ b/packages/puppeteer-extra-plugin-devtools/lib/RemoteDevTools.js
@@ -264,7 +264,7 @@ class DevToolsTunnel extends DevToolsCommon {
       debug('tunnel:close')
     })
 
-    tunnel.on('error', (err) => {
+    tunnel.on('error', err => {
       console.log('tunnel error', err)
     })
 
--- a/packages/puppeteer-extra-plugin-devtools/lib/RemoteDevTools.js
+++ b/packages/puppeteer-extra-plugin-devtools/lib/RemoteDevTools.js
@@ -265,7 +265,7 @@ class DevToolsTunnel extends DevToolsCommon {
     })
 
     tunnel.on('error', (err) => {
-      debug('tunnel errored', err)
+      console.log('tunnel error', err);
     })
 
     debug('tunnel:created', tunnel.url)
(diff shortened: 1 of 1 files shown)


berstend:
I'm wondering if we should use a `console.log` instead as a failing tunnel connection is quite removed from the happy path 😄 

PS: there's a global `uncaughtException` handler in Node.js that (when used surgically by silencing only specific issues) would help here as well, though it's better to hook into the tunnel error event.

PPS: I also noticed the localtunnel project added an intermediate page which I'm not a fan of, I guess for mission critical stuff a self-hosted instance might be preferable.

abowcut:
Was thinking to put a `console.log` at first but didn't want to inject console output if others didn't find it useful 😆
Added another commit with this change.

Thanks for pointing out the global `uncaughtException` handler, I hadn't seen/used it before. In this case seems better to handle the tunnel error event like you mentioned.

And regarding the localtunnel intermediate page, yes I've found it very annoying.
When I have a bit more time I'm planning to setup my own localtunnel/nginx server to avoid the intermediate page and get better reliability.
I noticed that swapping from using `https://localtunnel.me` to my own self-hosted instance will be easy with how you've setup the `DevToolsTunnel` class, so thanks for having the foresight to allow the configuration of the localtunnel host option 💯

Please let me know if there are any changes you want made in order to merge.
Thanks!

berstend:
image

Apologies for the nuisance but the CI linter is failing due to this small formatting issue, could you fix that? :)

abowcut:
Whoops sorry about that!
Should be fixed now.

I'm happy to run a linter before creating a pull request for any potential future prs.
Any suggestion for that?

berstend:
Thanks! It's possible for editors like VScode to pick up and use our linter/prettier config directly, alternatively `yarn lint` can be run in the plugin folder. :)

berstend:
```
Successfully published:
 - puppeteer-extra-plugin-devtools@2.3.3
```
thanks @abowcut 👍

Issue #428, opened 2021-02-13
[Bug] basic typescript integration fails with "no exported member" errors 


davidbielik:
**Describe the bug**
[Minimum Reproducible Code](https://github.com/berstend/puppeteer-extra/files/5975973/Archive.zip)

1. unzip this file
2. run `npm install`
3. run `tsc`

Expected: typescript compiles
Actual: lots of "namespace has no exported member" errors

This is just from importing the library. When I have actual puppeteer code i get many more errors for lots of other un-exported modules.

```
node_modules/puppeteer-extra-plugin/dist/index.d.ts:315:39 - error TS2694: Namespace '"/puppeteer-extra-bug/node_modules/puppeteer/lib/types"' has no exported member 'Worker'.

315     onWorkerCreated(worker: Puppeteer.Worker): Promise<void>;
                                          ~~~~~~

node_modules/puppeteer-extra/dist/puppeteer.d.ts:6:10 - error TS2305: Module '"../../puppeteer/lib/types"' has no exported member 'FetcherOptions'.

6 export { FetcherOptions } from 'puppeteer';
           ~~~~~~~~~~~~~~

[10:54:40 AM] Found 2 errors. Watching for file changes.
```

**Code Snippet**

```javascript
import puppeteer from "puppeteer-extra";
import StealthPlugin from "puppeteer-extra-plugin-stealth";
puppeteer.use(StealthPlugin());
```

**Versions**
```
System:
    OS: macOS 10.15.5
    CPU: (8) x64 Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
    Memory: 139.73 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 10.23.1 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.10 - /usr/local/bin/npm
  npmPackages:
    puppeteer: ^7.1.0 => 7.1.0 
    puppeteer-extra: ^3.1.16 => 3.1.16 
    puppeteer-extra-plugin-stealth: ^2.7.4 => 2.7.4 
```

berstend:
@davidbielik puppeteer broken typings, just look at this mess:
https://github.com/puppeteer/puppeteer/releases

Use `puppeteer@5` for now.

berstend:
I've added a note in the main readme (in the hopes people will see it):

[...]

PR #571, opened 2021-10-01
docs(puppeteer-extra): Remove types notice, add community plugin

claabs:
- Adds my plugin [puppeteer-extra-plugin-portal](https://github.com/claabs/puppeteer-extra-plugin-portal) to the community plugins section. I finally released it, so I figured I may as well add it to the list.
- Remove the puppeteer version notice for types now that #428 is closed


berstend (approved):
LGTM, thanks @claabs 👍

PR #592, opened 2021-11-23
feat(plugin-adblocker): Add cooperative intercept mode support

### LINKED ISSUE #364: Solving request interception chain

FdezRomero:
Adds support for the new [Cooperative Intercept Mode available in Puppeteer v11+](https://github.com/puppeteer/puppeteer/blob/v11.0.0/docs/api.md#upgrading-to-cooperative-mode-for-package-maintainers) and defining a custom `interceptResolutionPriority` for the following plugins:
- `puppeteer-extra-plugin-block-resources`
- `puppeteer-extra-plugin-block-adblocker`

This change checks that the required interception functions are available, so it's backwards compatible with the Legacy Mode and previous Puppeteer versions.

The `@cliqz/adblocker-puppeteer` dependency [already included support for Cooperative Mode in v1.23.0](https://github.com/cliqz-oss/adblocker/blob/master/packages/adblocker-puppeteer/CHANGELOG.md#v1230-mon-oct-18-2021).

Fixes #364, #156

### DIFF
--- a/packages/puppeteer-extra-plugin-adblocker/readme.md
+++ b/packages/puppeteer-extra-plugin-adblocker/readme.md
@@ -39,7 +39,10 @@ const puppeteer = require('puppeteer-extra')
 // Add adblocker plugin, which will transparently block ads in all pages you
 // create using puppeteer.
 const AdblockerPlugin = require('puppeteer-extra-plugin-adblocker')
-puppeteer.use(AdblockerPlugin())
+puppeteer.use(AdblockerPlugin({
+  // Optionally enable Cooperative Mode for several request interceptors
+  interceptResolutionPriority: 0
+}))
 
 // puppeteer usage as normal
 puppeteer.launch({ headless: true }).then(async (browser) => {
--- a/packages/puppeteer-extra-plugin-adblocker/readme.md
+++ b/packages/puppeteer-extra-plugin-adblocker/readme.md
@@ -38,10 +38,11 @@ const puppeteer = require('puppeteer-extra')
 
 // Add adblocker plugin, which will transparently block ads in all pages you
 // create using puppeteer.
+const { DEFAULT_INTERCEPT_RESOLUTION_PRIORITY } = require('puppeteer')
 const AdblockerPlugin = require('puppeteer-extra-plugin-adblocker')
 puppeteer.use(AdblockerPlugin({
   // Optionally enable Cooperative Mode for several request interceptors
-  interceptResolutionPriority: 0
+  interceptResolutionPriority: DEFAULT_INTERCEPT_RESOLUTION_PRIORITY
 }))
 
 // puppeteer usage as normal
--- a/packages/puppeteer-extra-plugin-adblocker/src/index.ts
+++ b/packages/puppeteer-extra-plugin-adblocker/src/index.ts
@@ -115,7 +115,7 @@ export class PuppeteerExtraPluginAdblocker extends PuppeteerExtraPlugin {
   /**
    * Sets the request interception priority on the `PuppeteerBlocker` instance.
    */
-   private setRequestInterceptionPriority(): void {
+  private setRequestInterceptionPriority(): void {
     this.blocker?.setRequestInterceptionPriority(this.opts.interceptResolutionPriority)
   }
 
--- a/packages/puppeteer-extra-plugin-adblocker/src/index.ts
+++ b/packages/puppeteer-extra-plugin-adblocker/src/index.ts
@@ -17,6 +17,8 @@ export interface PluginOptions {
   useCache: boolean
   /** Optional custom directory for adblocker cache files. Default: undefined */
   cacheDir?: string
+  /** Optional custom priority for interception resolution. Default: 0 */
+  interceptResolutionPriority?: number
 }
 
 /**
@@ -38,7 +40,8 @@ export class PuppeteerExtraPluginAdblocker extends PuppeteerExtraPlugin {
     return {
       blockTrackers: false,
       useCache: true,
-      cacheDir: undefined
+      cacheDir: undefined,
+      interceptResolutionPriority: 0
     }
   }
 
@@ -99,8 +102,10 @@ export class PuppeteerExtraPluginAdblocker extends PuppeteerExtraPlugin {
     if (this.blocker === undefined) {
       try {
         this.blocker = await this.loadFromCache()
+        this.blocker.setRequestInterceptionPriority(this.opts.interceptResolutionPriority)
       } catch (ex) {
         this.blocker = await this.loadFromRemote()
+        this.blocker.setRequestInterceptionPriority(this.opts.interceptResolutionPriority)
         await this.persistToCache(this.blocker)
       }
     }
--- a/packages/puppeteer-extra-plugin-adblocker/src/index.ts
+++ b/packages/puppeteer-extra-plugin-adblocker/src/index.ts
@@ -17,7 +17,7 @@ export interface PluginOptions {
   useCache: boolean
   /** Optional custom directory for adblocker cache files. Default: undefined */
   cacheDir?: string
-  /** Optional custom priority for interception resolution. Default: 0 */
+  /** Optional custom priority for interception resolution. Default: undefined */
   interceptResolutionPriority?: number
 }
 
@@ -41,7 +41,7 @@ export class PuppeteerExtraPluginAdblocker extends PuppeteerExtraPlugin {
       blockTrackers: false,
       useCache: true,
       cacheDir: undefined,
-      interceptResolutionPriority: 0
+      interceptResolutionPriority: undefined
     }
   }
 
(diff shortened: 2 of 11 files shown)


packages/puppeteer-extra-plugin-adblocker/src/index.ts
```diff
@@ -38,7 +40,8 @@ export class PuppeteerExtraPluginAdblocker extends PuppeteerExtraPlugin {
     return {
       blockTrackers: false,
       useCache: true,
-      cacheDir: undefined
+      cacheDir: undefined,
+      interceptResolutionPriority: 0
```
benallfree, line 44: I recommend defaulting to `undefined` to preserve legacy mode behavior. A priority should be explicitly set by the user if they want to use cooperative mode - at least, for now. Otherwise users may experience the handlers resolving in a different order than before, and that can be confusing.

tldr; Default to legacy mode and add a note to the README about how to enable cooperative mode for users who want it.

packages/puppeteer-extra-plugin-block-resources/index.js
```diff
@@ -65,7 +65,8 @@ class Plugin extends PuppeteerExtraPlugin {
         'other'
       ]),
       // Block nothing by default
-      blockedTypes: new Set([])
+      blockedTypes: new Set([]),
+      interceptResolutionPriority: 0
```
benallfree, line 69: Same note: `undefined` default

benallfree (changes requested):

packages/puppeteer-extra-plugin-block-resources/index.js
```diff
@@ -91,14 +92,50 @@ class Plugin extends PuppeteerExtraPlugin {
...
+      currentResolution,
+      currentPriority
+    })
+
+    if (alreadyHandled) return
+
+    return shouldBlock
+      ? request.abort('blockedbyclient', this.interceptResolutionPriority)
+      : request.continue(
+          request.continueRequestOverrides(),
```
benallfree, line 136: For backward compatibility, consider `request.continueRequestOverrides ? request:continueRequestOverrides() : undefined`. Same idea with `.abort()` and `.respond()`

benallfree (changes requested):

packages/puppeteer-extra-plugin-adblocker/src/index.ts
FdezRomero: Good point! I'll change it to be disabled by default 👍🏻

FdezRomero:
@benallfree I implemented your suggestions and fixes. I also included a TypeScript definition that I was using on my client app 😉 Let me know if you find something else.

packages/puppeteer-extra-plugin-block-resources/index.js
```diff
@@ -91,14 +94,57 @@ class Plugin extends PuppeteerExtraPlugin {
...
+      : true
+
+    this.debug('onRequest', {
+      type,
+      shouldBlock,
+      currentResolution,
+      currentPriority
+    })
+
+    if (alreadyHandled) return
```
benallfree, line 133: I think I missed this the first time. Can you walk me through the intent? In cooperative mode, the Puppeteer core will run all handlers and resolve the intercept to the one with the highest priority. Unless you have a special case, your handler should simply do its computation and respond at whatever priority it is set to. If a different response at a higher priority takes place, it will prevail and yours will “lose”. If yours is the highest priority response, it will “win”

benallfree (changes requested):

packages/puppeteer-extra-plugin-block-resources/index.js
FdezRomero: @benallfree The problem is that if you use a plugin that hasn't been migrated to Cooperative Mode and still uses Legacy Mode, the request is handled right away.

So if we call to `continue`/`abort`/`respond` etc. then we get the `Request is already handled!` error again (I just tried it again to verify). That's why we need this condition here, because we can't asume all plugins are opting-in to the new mode.

benallfree: Oh that’s very smart, I think I’ll suggest this pattern in the official docs! Well done.

benallfree (approved):

FdezRomero:
@benallfree Glad to hear that! It seems that this PR needs a second approval. Do you know who can help with that? Thanks!

benallfree:
@FdezRomero I'm not a maintainer of this package so I'm not sure how to help with a 2nd approval.

By the way, I just submitted https://github.com/puppeteer/puppeteer/pull/7795/files based on your work here. It fixes that spelling mistake and officially exposes `isInterceptResolutionHandled()` logic. You might consider waiting on this merge until we know whether the Puppeteer team will accept my PR.

benallfree:
@FdezRomero Small note, cooperative support came in 10.2 https://github.com/puppeteer/puppeteer/blob/v11.0.0/CHANGELOG.md#1020-2021-08-04

benallfree:
@FdezRomero What will break?  #7796 does not introduce breaking changes, or it’s not supposed to :)

FdezRomero:
You're right @benallfree, `@cliqz/adblocker-puppeteer` isn't using the `interceptResolution()` method so it won't be affected by the rename to `interceptResolutionState()`. It doesn't use it to check if the request has been already handled or not 👍🏻

I moved this PR back to draft and will apply the changes once https://github.com/puppeteer/puppeteer/pull/7796 ships. Hopefully it gets reviewed soon!.

benallfree:
@FdezRomero  puppeteer/puppeteer#7796 has been accepted by the Puppeteer team 👍

FdezRomero:
@benallfree Great! I'll update the PR with the new API 👍🏻
Any idea when a new version might be released?

benallfree:
@FdezRomero No idea 😅

FdezRomero:
Hi @benallfree, I have updated `adblocker-puppeteer` to the latest version with the fixes and upgraded the code to your newer Puppeteer API, which also made the code easier to read. Would you mind giving it another look? Thanks! 🙏🏻

packages/puppeteer-extra-plugin-block-resources/index.js
```diff
@@ -91,14 +94,50 @@ class Plugin extends PuppeteerExtraPlugin {
...
+    this.debug('onRequest', {
+      type,
+      shouldBlock,
+      alreadyHandled
+    })
+
+    if (alreadyHandled) return
+
+    if (shouldBlock) {
+      const abortArgs = request.continueRequestOverrides
```
benallfree, line 129: Are you using `continueRequestOverrides` purely for version detection? If so, it might make more sense to use  [abortErrorReason](https://github.com/puppeteer/puppeteer/blob/dc23b7535cb958c00d1eecfe85b4ee26e52e2e39/src/common/HTTPRequest.ts#L220) since this is an `abort()` call.

packages/puppeteer-extra-plugin-adblocker/src/index.ts
```diff
@@ -99,14 +102,23 @@ export class PuppeteerExtraPluginAdblocker extends PuppeteerExtraPlugin {
...
       }
     }
     return this.blocker
   }
 
+  /**
+   * Sets the request interception priority on the `PuppeteerBlocker` instance.
+   */
+  private setRequestInterceptionPriority(): void {
+    this.blocker?.setRequestInterceptionPriority(this.opts.interceptResolutionPriority)
```
benallfree, line 119: Not familiar enough with the lib to know if this is correct, it seems to be passing it down a layer?

benallfree (approved):
lgtm w optional suggestions 👍

packages/puppeteer-extra-plugin-adblocker/src/index.ts
FdezRomero: @benallfree Yes exactly, it references the instance of `adblocker-puppeteer` that is used internally, which is the one that really intercepts and handles the requests, and this `pupeteer-extra` plugin acts as a wrapper: https://github.com/ghostery/adblocker/blob/master/packages/adblocker-puppeteer/adblocker.ts#L282-L283

packages/puppeteer-extra-plugin-adblocker/readme.md
```diff
@@ -39,7 +39,10 @@ const puppeteer = require('puppeteer-extra')
 // Add adblocker plugin, which will transparently block ads in all pages you
 // create using puppeteer.
 const AdblockerPlugin = require('puppeteer-extra-plugin-adblocker')
-puppeteer.use(AdblockerPlugin())
+puppeteer.use(AdblockerPlugin({
+  // Optionally enable Cooperative Mode for several request interceptors
+  interceptResolutionPriority: 0
```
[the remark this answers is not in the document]
FdezRomero: Yes, it's exported. I changed it in all examples 👍🏻

packages/puppeteer-extra-plugin-block-resources/index.js
FdezRomero: I didn't know about this method, thanks! I changed it for the feature detection 👍🏻

FdezRomero:
I think it's good to go! Let's see if it can get a bit of love from the maintainers 🤞🏻😄 @berstend

FdezRomero:
Happy new year @berstend! 🎉

How can we get a second review to get this merged? We already made the necessary changes upstream in Puppeteer and Adblocker to support Cooperative Intercept mode in `puppeteer-extra` 😉

Thanks!

benallfree (approved):

deldrid1 (approved):

deldrid1:
@berstend - Any chance this can be reviewed? This is a critical feature I need!

benallfree:
@deldrid1 @berstend You both might consider running your own fork for a while, it seems the maintainer is busy on other things.

berstend:
```
Successfully published:
 - puppeteer-extra-plugin-adblocker@2.13.0
 - puppeteer-extra-plugin-block-resources@2.4.0
 ```

PR #615, opened 2022-01-26
Append patch version to UA hint platform version

momala454:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-CH-UA-Platform-Version
Sec-CH-UA-Platform-Version contains `dot-separated major, minor and patch version numbers`
The user agent may not contains 3 numbers, example on windows 10 it contains Windows 10.0, so we add the patch version at the end.

Take note that on linux the version seems to contains the kernel version, which this evasion doesn't get so if you set `maskLinux` to false on linux, the platformVersion will be empty while it shouldn't.

On android, on my phone on android 11, there is no header `Sec-CH-UA-Platform-Version` so i don't know if it must be removed

The bitness is also not filled.

### DIFF
--- a/packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.js
+++ b/packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.js
@@ -129,14 +129,16 @@ class Plugin extends PuppeteerExtraPlugin {
       return greasedBrandVersionList
     }
 
+	const _padPlatformVersion = (version) => version+((version.split('.').length<=2)?'.0':'')
+
     // Return OS version
     const _getPlatformVersion = () => {
       if (ua.includes('Mac OS X ')) {
-        return ua.match(/Mac OS X ([^)]+)/)[1]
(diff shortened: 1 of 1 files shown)


Niek (approved):

Niek:
Approved, ping @berstend 

Relevant URL: https://chromestatus.com/feature/5080939765956608

packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.js
```diff
@@ -129,14 +129,16 @@ class Plugin extends PuppeteerExtraPlugin {
       return greasedBrandVersionList
     }
 
+	const _padPlatformVersion = (version) => version+((version.split('.').length<=2)?'.0':'')
```
ODGodinho, line 132: format the code for better readability
```suggestion
    const _padPlatformVersion = (version) => version + ((version.split('.').length <= 2) ? '.0' : '')
```

PR #640, opened 2022-05-27
fix(plugin-recaptcha): Update hcaptcha challenge query selector

jonas-duarte:
The query selector from the method `_findActiveChallenges` changed, so I updated it.

However, I didn't find a good way to test this, because the `test('will detect hCAPTCHAs')` uses [this url](http://democaptcha.com/demo-form-eng/hcaptcha.html) and it is tested by the method `_findRegularCheckboxes`.


berstend:
Can confirm it changed, on way to test this it to go to https://democaptcha.com/demo-form-eng/hcaptcha.html and issue ` hcaptcha.execute()` 😄

Issue #646, opened 2022-06-03
Ready for a new Typescript upgrade ?


UrielCh:
Hi after a year away from this project, I'm back.
I'm happy that my previous PRs (#455 #461) had help in #555.

My next change will convert more remaining JS files to TS.

To avoid applying many changes, I would like to know the official way to configure puppeteer-extra-plugin-stealth's evasions options?

For example, 'stealth/evasions/navigator.hardwareConcurrency' accept a `hardwareConcurrency` options, but I did not find any doc about providing such a value.

UrielCh:
Your project can not be built on a Windows host Can I fix that too?

UrielCh:
I will drop TsLint, ESLint should be used instead.

UrielCh:
Full change list is available in #649

berstend:
Hey, thanks for the PR - it's indeed massive 😄 I need a moment to go through the changes, but it's definitely great as reference and potential backlog.

We have a larger change on the horizon for the stealth evasions: We need to collect their scripts first and only inject them once to help with a set of issues, so some larger changes will happen here which I need to clarify first.

UrielCh:
Tell me what do you need for a quick merge, I need to use my fork quickly.
Or I Will need to publish a forked version supporting alternative naming conversion.

Can I drop some old pptr version support?
I would like to support only pptr 8+
So I can simplify typing injection.

UrielCh:
I see that you add some:
` "types": "index.d.ts",`
VS my 
  "typings": "index.d.ts",

I can read [here](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html)
```
Note that the "typings" field is synonymous with types, and could be used as well.

Also note that if your main declaration file is named index.d.ts and lives at the root of the package (next to index.js) you do not need to mark the types property, though it is advisable to do so.
```

UrielCh:
you gave me extra work with your pretty useful last changes :wink:

can you explain me this change:
`const client =  page && typeof page._client === 'function' ? page._client() : page._client`

Which version of pptr has _client as a function and which one as a field?

berstend:
Since puppeteer@14.4.1: page._client changed to page._client()

berstend:
> Tell me what do you need for a quick merge, I need to use my fork quickly.
> Or I Will need to add a support custom plugin naming to use a forked version.

I don't think I can merge that quickly - what's the most pressing thing you need that makes you consider forking?

[...]

PR #455, opened 2021-03-18
upgrade code to support puppeteer 8.0.0

UrielCh:
puppeteer 8.0.0 is writen in typescript, so we do not need @types/puppeteer

puppeteer use some typescript v4+ feature, I upgrade typescript version to the same as puppeteer  currently use.

do you realy want to keep retrocompatibility with old puppeteer ?

if to the gitflow must be change. (multiple option are posible here)

I tryed to made as little change as possible.

@cliqz/adblocker-puppeteer must be upgrade to 1.20.+
typescript must be 4+, ^4.1.5 is currently use in puppeter source.

let me know you point.

PR #649, opened 2022-06-10
Big puppeteer-extra upgrade.

UrielCh:
One of my Biggest PR.

This PR contains the following changes:

- Mostly typescript
- Dependency loader rewrite.
`_getMissingDependencies(plugins: any)` removed in favor of `get dependencies()`
- Enforce typing
- Works on windows
- using default export everywhere
```typescript
export default (pluginConfig?: Partial<PluginOptions>) =>new Plugin(pluginConfig)
```
- puppeteer-extra loader can inject option to plugins
```typescript
  PuppeteerExtra.setPluginOptions(name: PluginName, options: any): void {
```
- puppeteer-extra-plugin-stealth loader can inject option to evasions

```typescript
puppeteer.use(plugin({
  evasionsOptions: {
    "navigator.languages": { languages : ["es-ES", "es"]},
    "navigator.hardwareConcurrency": { hardwareConcurrency: 12 },
 }}));
```
in this case typing enforce all Plugin constuctors.
- fix invalid `interface PluginData`

and many more changes.


prescience-data:
I believe this came up last time around, but you will find it is a lot more "friendly" to do many small, tightly scoped PRs rather than forcing a maintainer to reason through a huge one.

The reason for this is that it is far more likely to catch bugs when the code is small "easy-to-digest" pieces, so large PRs are often not accepted due to risks of this that increase exponentially with size. 

Or to put it visually 🤭: 

image

prescience-data:
One immediate thing I note is: 

> "using default export everywhere"

I'm generally not a fan of this as default exports make it impossible for IDE helpers to handle refactoring magic when something changes.


### With explicit exports

```typescript
// foo.ts
export const bar: string = "bar"
```

```typescript
// This allows the IDE to auto-refactor variable name...
import { bar } from "./foo"
```

### With default exports
```typescript
// foo.ts
export default bar: string = "bar"
```

```typescript
// The IDE will not be able to refactor "bar" in this case...
import bar from "./foo"
```

UrielCh:
I open issue #646 before starting, but I did not get any comment.
so I was not sure to get an answer to this PR.

And this is not my first big PR in this project see #455 and #461 

## About the default import:
Default export are available but not mandatory, all plugin contains at least 3 export:
- export interface PluginOptions
- export class Plugin
- export default (pluginConfig?: Partial<PluginOptions>) => new Plugin(pluginConfig)

The loading code is now:
```typescript
          const req = require(fullname);
          const shortName: PluginName = fullname.substring('puppeteer-extra-plugin-'.length);
          const opts = this._pluginOptions[shortName] || undefined;
          let dep: PuppeteerExtraPlugin | null = null
          // Try to require and instantiate the stated dependency
          if ('default' in req) {
            dep = req.default(opts)
          } else {
            dep = req(opts)
          }
```
I use a default export, but the Plugin export is also available
so a code like that will also work:

```
          if ('Plugin' in req) {
            dep = new Plugin(opts)
          } else  if ('default' in req) {
            dep = req.default(opts)
          } else {
            dep = req(opts)
          }
```
The named exporter Plugin is a Class, the default export is a function.

UrielCh:
Most of the added lines are coming from 
puppeteer-extra-plugin-stealth/examples/selfthosted.ts
puppeteer-extra-plugin-stealth/examples/files
contains a copy of 
- https://intoli.com/blog/not-possible-to-block-chrome-headless/chrome-headless-test.html
- https://bot.sannysoft.com/

UrielCh:
PR completed ready for review.

UrielCh:
I have just converted my puppeteer jquery module to puppeteer-extra-plugin
[puppeteer-extra-plugin-jquery](https://www.npmjs.com/package/puppeteer-extra-plugin-jquery)

I have some more changes to push to puppeteer-extra for the next PR.

berstend:
@UrielCh I've published `playwright-extra` in #664 

It contains new/refactored code, clean deps, new test runners and a few other features like this one:
```js
import { chromium } from 'playwright-extra'

import StealthPlugin from 'puppeteer-extra-plugin-stealth'
chromium.use(StealthPlugin())

// New way to overwrite the default options of stealth evasion plugins
// https://github.com/berstend/puppeteer-extra/tree/master/packages/puppeteer-extra-plugin-stealth/evasions
chromium.plugins.setDependencyDefaults('stealth/evasions/webgl.vendor', {
  vendor: 'Bob',
  renderer: 'Alice'
})
```

Once the dust has settled doing housekeeping in the repo (especially updating dependencies/linting, etc) is in order 😄 

I used your PR as a reference while implementing the new code 👍

UrielCh:
Hi,
Can you integrate my `"ambient-dts-fix-path": "node ../post-replace.js"`
and drop replace-in-files; your `replace-in-files` are platform dependent.

ppyrzanowski:
Any updates on this pull request? Windows compatibility is still not available, and proper typescript rewrites to the plugins and rest would improve overall dev experience. This PR would resolve both of my main pain points.

PR #654, opened 2022-06-18
bugfix: puppeteer@14.4.1: page._client -> page._client()

peterlevel1:
packages/puppeteer-extra-plugin-stealth/evasions/user-agent-override/index.js:180


berstend:
Thanks! As mentioned #653 is more complete so closing this PR in favor of this one

PR #657, opened 2022-06-22
fix(plugin-stealth): Revert iframe.contentWindow changes and add tests

berstend:
Unfortunately the changes in #565 would break captcha frames, I reverted the `iframe.contentWindow` changes until we had time to improve the new code.

I've also added/improved 3 captcha specific tests that make sure we don't break them in the future :)

@regseb if you want to improve the code feel free to do so, I left in your (now unused) utils additions and commented out the new tests that would break


berstend:
```
Successfully published:
 - puppeteer-extra-plugin-stealth@2.10.1
 ```

PR #664, opened 2022-07-03
feat: Add playwright-extra

berstend:
This PR:
- adds a new `playwright-extra` package
- it features a puppeteer compatibility layer so existing plugins derived from `PuppeteerExtraPlugin` (and expecting certain puppeteer methods) can be used with it
- currently tested and optimized for the stealth (chromium) and recaptcha (chromium, firefox, webkit) plugin

This approach is much more sane as opposed to switching everything to a new shared plugin base class at once, as attempted in #303 
Further background on the reasoning behind this strategy change can be found here: https://github.com/berstend/puppeteer-extra/issues/454#issuecomment-1169678855

The code (informed by maintaining a plugin framework for 4 years now 😄) is in parts much more elegant than puppeteer-extra and once some overdue housekeeping in the repo has been done I plan to refactor puppeteer-extra as well

The next steps are
- fixing any bugs that might come up related to the core framework functionality (augmenting playwright, plugin handling/events)
- improving the compatibility layer and supporting more puppeteer-extra plugins
- make very slight changes to the existing plugins to help improve playwright usage (e.g. the Page type mods in the recaptcha plugin)
- more documentation (new features, playwright caveats, how to contribute playwright plugins)

Eventually
- deciding on the best way to properly support both playwright & puppeteer (most likely with a new shared base plugin class)
- code reuse between playwright-extra & puppeteer-extra

PR #665, opened 2022-07-03
feat(plugin-recaptcha): Add playwright type mods

berstend:
This allows e.g. `await page.solveRecaptchas()` to be used with type support when playwright-extra is driving the plugin (#664)


amiralis1365:
Throws Typescript error when you install it with puppeteer and playwright-core is absent. `Cannot find module 'playwright-core' or its corresponding type declarations.`

PR #674, opened 2022-07-07
chore:(playwright-extra): Bump minimum Node version to 12

### LINKED ISSUE #670: [Bug] Package playwright-extra v4.3 breaks on node v10
Hi ppl,

The node engine version for the playwright-extra is set to 8:
https://github.com/berstend/puppeteer-extra/blob/5b39abb432e3d86675a899f3a5ba351d11b7b7ba/packages/playwright-extra/package.json#L38-L40

but new code in the v4.3 breaks with node v10.
One example of breaking code is:

https://github.com/berstend/puppeteer-extra/blob/5b39abb432e3d86675a899f3a5ba351d11b7b7ba/packages/playwright-extra/src/helper/loader.ts#L19

It uses the Object.fromEntries that is only available in node v12.

I observed that the github workflows use the node v14 as its minimum version.
Maybe its better to upgrade all node engine versions to "node": ">=14"?

Only for reference, our application crashed after upgrading from the v4.2.1-next.616 (now deprecated) to the v4.3. 
It is a legacy application using node v10.

berstend:
Fixes #670

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,8 +26,7 @@ jobs:
     strategy:
       matrix:
         node:
-          - 16
-          # - 14
+          # - 16
           - 14
         puppeteer_version:
           # - 14.2.0 # Chromium 103.0.5059.0 # requires >=14.1.0
--- a/packages/playwright-extra/package.json
+++ b/packages/playwright-extra/package.json
@@ -36,7 +36,7 @@
     "pupeteer"
   ],
   "engines": {
-    "node": ">=8"
+    "node": ">=12"
   },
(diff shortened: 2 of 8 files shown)


berstend:
```
Successfully published:
 - extract-stealth-evasions@2.6.3
 - playwright-extra@4.3.1
 - puppeteer-extra-plugin-anonymize-ua@2.4.2
 - puppeteer-extra-plugin-devtools@2.4.2
 - puppeteer-extra-plugin-recaptcha@3.6.4
 - puppeteer-extra@3.3.2
 ```

PR #698, opened 2022-08-16
fix typo :)

jackdewan:


9je (changes requested):
extremely minor but important change

ushuz (approved):

PR #701, opened 2022-08-28
feat(plugin-recaptcha) Allow custom data to be sent

bert-w:
An improvement to previous additions that adds the `extraData` option as a parameter. Note: I did not test this code since I dont currently have the typescript build env installed.

This change should be backwards compatible (optional argument) and is a dirty fix to allow developers to send anything to the provider (there are some 2captcha API options that I could otherwise not send).

### DIFF
--- a/packages/puppeteer-extra-plugin-recaptcha/src/index.ts
+++ b/packages/puppeteer-extra-plugin-recaptcha/src/index.ts
@@ -195,7 +195,8 @@ export class PuppeteerExtraPluginRecaptcha extends PuppeteerExtraPlugin {
 
   async getRecaptchaSolutions(
     captchas: types.CaptchaInfo[],
-    provider?: types.SolutionProvider
+    provider?: types.SolutionProvider,
+    extraData: { [key: string]: any } = {},
   ) {
     this.debug('getRecaptchaSolutions', { captchaNum: captchas.length })
     provider = provider || this.opts.provider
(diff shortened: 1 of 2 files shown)


packages/puppeteer-extra-plugin-recaptcha/src/index.ts
```diff
@@ -331,7 +334,7 @@ export class PuppeteerExtraPluginRecaptcha extends PuppeteerExtraPlugin {
     prop.enterRecaptchaSolutions = async (solutions: types.CaptchaSolution[]) =>
       this.enterRecaptchaSolutions(prop, solutions)
     // Add convenience methods that wraps all others
-    prop.solveRecaptchas = async () => this.solveRecaptchas(prop)
+    prop.solveRecaptchas = this.solveRecaptchas.bind(this, prop)
```
bert-w, line 337: i dont know if this syntax works with `async`. `bind()` here makes sure that the first parameter is bound to the page object. Further calls to `prop.solveRecaptchas` should then call the `solveRecaptchas` with just 1 argument (the `extraData`) since the `page` has been bound thus disappeared

bert-w: tested this locally; works correctly (and properly assigns the function so that any additional arguments in the future will also work)

AndresGaibor (approved):

Issue #740, opened 2022-10-26
[Bug] `PuppeteerExtraPluginRecaptcha` error has no information and cannot be evaluated


markcellus:
**Describe the bug**

Currently, when using the `puppeteer-extra-plugin-recaptcha` package with `throwOnError` set to `true` and a `PuppeteerExtraPluginRecaptcha` error is thrown [here](https://github.com/berstend/puppeteer-extra/blob/523b2efe7dc000536ef6bd608810d4e01e206504/packages/puppeteer-extra-plugin-recaptcha/src/index.ts#L191), the argument passed to `Error` is an object. So the information on the error gets lost, leaving no way for consumer code to know about what caused the error. 

It just outputs the following as a console warning:

```bash
PuppeteerExtraPluginRecaptcha: An error occured during "getRecaptchaSolutions": {
_vendor: 'recaptcha',
provider: '2captcha',
id: 'mwx6k8wq6yxx',
requestAt: 2021-05-09T01:35:25.450Z,
error: 'Error: 2captcha error: [redacted]'
}
```

...and the thrown error is an empty object, so it can't be evaluated:

```js
try {
        let { captchas } = await page.findRecaptchas();
        if (!captchas.length) {
            return;
        }
        await page.solveRecaptchas();
    } catch (e) {
        console.log(e) // <== cannot evaluate this error because it has no information
    }
```

markcellus:
I've opened https://github.com/berstend/puppeteer-extra/pull/741 to attempt to address. Open to any feedback! Thanks!

PR #741, opened 2022-10-26
Expose `PuppeteerExtraPluginRecaptcha` error

markcellus:
Resolves https://github.com/berstend/puppeteer-extra/issues/740 by updating the `PuppeteerExtraPluginRecaptcha` thrown error and exposes all of its properties to be evaluated by consumer code.

### DIFF
--- a/packages/puppeteer-extra-plugin-recaptcha/src/index.ts
+++ b/packages/puppeteer-extra-plugin-recaptcha/src/index.ts
@@ -235,7 +235,16 @@ export class PuppeteerExtraPluginRecaptcha extends PuppeteerExtraPlugin {
       )
     }
     if (this.opts.throwOnError && response.error) {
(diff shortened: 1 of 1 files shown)


markcellus:
Perhaps a better way to do this is to create a new error class that extends `Error` (e.g. `PuppeteerExtraPluginRecaptchaError`), but I'm not too sure whether that would be acceptable here.

berstend:
Yeah, the error handling isn't optimal 😄 
From a quick look it seems that a `response.error` can be a string or an Error (in which case wrapping it again in an error would cause the info to get lost)

I wonder if this would work as a fix as well:
```js
if (this.opts.throwOnError && response.error) {
  throw !!response.error?.message ? throw response.error : throw new Error(response.error)
}
```

PR #745, opened 2022-11-01
Resolved executablePath issue from the example.

vikasg603:
In the latest version of puppeteer-core, it is mandatory to pass executablePath, or else it throws the below error 
```bash
Error: An `executablePath` or `channel` must be specified for `puppeteer-core`
```

This pull request will resolve this issue from the example code.

### DIFF
--- a/packages/puppeteer-extra/readme.md
+++ b/packages/puppeteer-extra/readme.md
@@ -22,6 +22,7 @@ yarn add puppeteer@2.0.0 puppeteer-extra
 // it augments the installed puppeteer with plugin functionality.
 // Any number of plugins can be added through `puppeteer.use()`
 const puppeteer = require('puppeteer-extra')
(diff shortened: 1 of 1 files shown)


AhmadIbrahiim:
With puppeteer `19.x.x` this should work for you.
```js
import { executablePath } from 'puppeteer';
import puppeteer from 'puppeteer-extra';

  const browser = await puppeteer.launch({
    headless: true,
    executablePath: executablePath(),
  });
  
```

vikasg603:
But @AhmadIbrahiim don't you think this should also be an part of the docs?

AhmadIbrahiim:
> But @AhmadIbrahiim don't you think this should also be an part of the docs?

Yeah, It should be

AhmadIbrahiim (approved):
I think this looks really good!

PR #755, opened 2022-11-16
fix(puppeteer-extra): Prefer puppeteer over puppeteer-core

ushuz:
puppeteer started to depend on puppeteer-core since probably [v18.1.0](https://github.com/puppeteer/puppeteer/blob/bafdb4784ccfe4a35fec9e6d366e4d1b40a82e1b/packages/puppeteer/package.json), but there are some API limitations on puppeteer core, like `puppeeter.defaultDownloadPath()`


berstend:
Thanks!
```
Successfully published:
puppeteer-extra@3.3.6
```

PR #758, opened 2022-11-28
fix(plugin-recaptcha): Support esbuild transpilation of content scripts

igalklebanov:
This one was a real pain in the *@#^@.
But now the sun can shine once again, and the scraping can continue!

Basically, was having a blast running `plugin-recaptcha` locally with `ts-node`, and once the same code hit production aws lambdas it started throwing the mentioned error.

Long story short, it was `esbuild` and the way it transformed `RecaptchaContentScript` & `HcaptchaContentScript` classes into something like `var HcaptchaContentScript = class {`.

The fix looks for such a sequence and injects the class name back.

### DIFF
--- a/packages/puppeteer-extra-plugin-recaptcha/src/index.ts
+++ b/packages/puppeteer-extra-plugin-recaptcha/src/index.ts
@@ -72,6 +72,11 @@ export class PuppeteerExtraPluginRecaptcha extends PuppeteerExtraPlugin {
       scriptSource = HcaptchaContentScript.toString()
       scriptName = 'HcaptchaContentScript'
     }
+    // Some bundlers transform classes to anonymous classes that are assigned to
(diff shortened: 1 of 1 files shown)


transitive-bullshit:
Also running into this on https://github.com/transitive-bullshit/chatgpt-api

I ended up switching captcha providers because it wouldn't work when invoked via `tsx` (which uses `esbuild` under the hood).

igalklebanov:
I think I tried `tsx` too, but either it didn't work because of this issue or something with decorators. So went with `ts-node` because well, build is not as slow as waiting for hcaptcha provider's response.

berstend:
@igalklebanov great find 👍 

I'll try to merge this in before merging #763 for attribution :)

berstend:
```
Successfully published:
 - puppeteer-extra-plugin-recaptcha@3.6.8
```

Issue #459, opened 2021-03-22
[Proposal] Session persistence using cookies, local/session storage and indexedDB stores


Shisuki:
### Plugin Name: Session Persist

I wanna add a new plugin that persists the session when get/set cookies fail, recently I stumbled on an issue regarding a website that uses redux for session persistence and the only way to keep the session active for later use was to save all the cookies using CDP (get all cookies from all domains, but might work with just a simple cookie get/set), get local and session storage and the indexedDB stores.

For this to be effective those need to be saved on each frameNavigated (url change), it doesn't seem to be a big overhead but if you have any other suggestions I'm listening.

To set the cookies, what i did is navigate to the domain first, clear everything, set our previously saved values, then navigate to the target url.

I'm ready to work on it and follow the coding guidelines

berstend:
I like the idea of a plugin like this 😄 

Looking at Devtools/Application:
![image](https://user-images.githubusercontent.com/1368633/112298047-df0baf00-8c96-11eb-8a41-8020d0c8a3e5.png)

So the plugin would be able to save/restore everything under Storage right?

> To set the cookies, what i did is navigate to the domain first, clear everything, set our previously saved values, then navigate to the target url.

Without have looked closer at the CDP methods I'm wondering if there's a way to avoid that page navigation.

In terms of user-facing API what would be most flexible could be this?
```js
const sessionJSON = await session.save()
await session.restore(sessionJSON)
```

berstend:
For reference: https://chromedevtools.github.io/devtools-protocol/tot/Storage/#method-getCookies

![image](https://user-images.githubusercontent.com/1368633/112299479-1fb7f800-8c98-11eb-983c-83de67b5a81f.png)

clouedoc:
I was also planning on creating that kind of plugin!

### About IndexedDB persistence

On some targets you need to recreate the database completely and edit the schema.
This implies saving the version of the database.
Reference: https://stackoverflow.com/questions/46541435/check-and-increment-the-version-number-of-an-indexeddb

clouedoc:
I've uploaded a [first draft](https://github.com/clouedoc/puppeteer-extra-plugin-session) yesterday. The code is there, it only needs a few modifications and being wired up in an easy-to-use plugin.

I would like the API to look like this:
```typescript
puppeteer.use(SessionPlugin()); 
// [...]
await page.goto('https://github.com')
// [...]
[...]

PR #804, opened 2023-05-07
Persist cookies + localStorage plugin

kangoo13:
Hello everybody,

The ways to retrieve new cookies/localStorage are a bit tricky but I didn't find any other better solution, yet.

If you have any idea do not hesitate I can implement them.

The plugin is available here meanwhile the PR is merged : 
https://www.npmjs.com/package/puppeteer-extra-plugin-session-persistence

This is my first code on TypeScript I am not really used to it, do not hesitate to tell me how can I improve this code !

#459


clouedoc (commented):
If you want, here is some code for IndexedDB: https://github.com/clouedoc/puppeteer-extra-plugin-session/blob/master/src/providers/indexedDb.ts

1. Do not use tslint; use ESLint with typescript-eslint instead. This is because TSLint is now deprecated. Rush Stack has a great starter ESLint file.
2. Polling might be a bit aggressive
3. IIRC NodeJS 20 has a built-in testing tool
4. Storage engines feel out of scope. The responsibility is better left to the user, or there will be complaints in GitHub issues. Of course, that's only my opinion; I appreciate the value of having storage engines available out-of-the-shelf 🙂 

> todo: make a PR to puppeteer-extra to change the PluginData interface, name and value should not be an object but any instead.

No, it should be unknown to force the developer to do a type assertion.`any`s are to be avoided at all costs. Every time you commit an `any`, a developer at the end of the world gets depression.
To learn this kind of stuff, I highly recommend the book _Efficient TypeScript_

In this specific case, you should have created an interface with the structure of how your plugin options should look like.

---

In any case, enormous props for submitting your code 🙂 
You are very welcome in our small community 👍

PR #913, opened 2024-11-25
Changed @cliqz/adblocker-puppeteer with @ghostery/adblocker-puppeteer

dave9123:
image
[@cliqz/adblocker-puppeteer](https://npmjs.com/package/@cliqz/adblocker-puppeteer) is now depreciated


Tyson-core:
The error in the job logs is due to an unexpected type for the label 'package: core'. The configuration file `.github/labeler.yml` expects the label to be an array of configuration options but found a different type instead.

Steps to Fix the Error:
Check the `.github/labeler.yml `Configuration:
Ensure that all labels specified in `.github/labeler.yml `are arrays of configuration options.

Correct the Label Configurations:
For each label in the configuration file, make sure it follows the correct format. For example:

```
YAML
package:
  - core
```

